PR #63, opened 2020-03-16
Added support for multiple application

vinodkumar3:


kanika1894:
The final placing of workingDirectory field needs to be decided. Rest all looks good.

package.json
```diff
@@ -2,7 +2,7 @@
     "name": "azure-deploy",
     "displayName": "Deploy to Azure",
     "description": "Generating CI/CD pipelines to Azure from GitHub and Azure Repo",
-    "version": "1.0.4",
```
bishal-pdMSFT: lets make this change only we are about to ship a newer version of extension. Otherwise no need to update.

src/configure/configure.ts
```diff
@@ -443,16 +443,41 @@ class Orchestrator {
...
 
         telemetryHelper.setTelemetry(TelemetryKeys.ChosenTemplate, this.inputs.pipelineConfiguration.template.label);
     }
 
+    private async updateRepositoryAnalysisApplicationSettings(repoAnalysisResult: RepositoryAnalysisParameters): Promise<void>{
+        var applicationSettings = repoAnalysisResult.repositoryAnalysisApplicationSettingsList.filter(applicationSetting => {
+            return applicationSetting.language === this.inputs.pipelineConfiguration.template.language;
+        });
+        
+        let workspacePaths = Array.from(new Set(applicationSettings.map(a => a.settings.workingDirectory)));
```
bishal-pdMSFT: `applicationSettings.map(a => a.settings.workingDirectory)` should itself return an array. Why do you need to do `Array.from(new Set(...))` ?

src/configure/configure.ts
```diff
@@ -443,16 +443,41 @@ class Orchestrator {
...
-                return applicationSettings.language === this.inputs.pipelineConfiguration.template.language;
-            });
+            this.updateRepositoryAnalysisApplicationSettings(repoAnalysisResult);
         }
 
         telemetryHelper.setTelemetry(TelemetryKeys.ChosenTemplate, this.inputs.pipelineConfiguration.template.label);
     }
 
+    private async updateRepositoryAnalysisApplicationSettings(repoAnalysisResult: RepositoryAnalysisParameters): Promise<void>{
+        var applicationSettings = repoAnalysisResult.repositoryAnalysisApplicationSettingsList.filter(applicationSetting => {
```
bishal-pdMSFT: You should also handle the case where no applicationSetting matches and then trace an error

src/configure/configure.ts
```diff
@@ -443,16 +443,41 @@ class Orchestrator {
...
+        for (let workspacePath of workspacePaths) {
+            workspacePathQuickPickItemList.push({ label: workspacePath, data: workspacePath });
+        }
+        let selectedWorkspacePathItem = await this.controlProvider.showQuickPick(
+            constants.SelectWorkspace,
+            workspacePathQuickPickItemList,
+            { placeHolder: Messages.selectWorkspace });
+        
+        this.inputs.pipelineConfiguration.workingDirectory = selectedWorkspacePathItem.data;
+        this.inputs.repositoryAnalysisApplicationSettings = repoAnalysisResult.repositoryAnalysisApplicationSettingsList.find(applicationSettings => {
```
bishal-pdMSFT: too long lines. please format into multiple lines

src/configure/configure.ts
```diff
@@ -443,16 +443,41 @@ class Orchestrator {
...
-            this.inputs.repositoryAnalysisApplicationSettings = repoAnalysisResult.repositoryAnalysisApplicationSettingsList.find(applicationSettings => {
-                return applicationSettings.language === this.inputs.pipelineConfiguration.template.language;
-            });
+            this.updateRepositoryAnalysisApplicationSettings(repoAnalysisResult);
         }
 
         telemetryHelper.setTelemetry(TelemetryKeys.ChosenTemplate, this.inputs.pipelineConfiguration.template.label);
     }
 
+    private async updateRepositoryAnalysisApplicationSettings(repoAnalysisResult: RepositoryAnalysisParameters): Promise<void>{
```
bishal-pdMSFT: rename to `selectWorkingDirectory`

src/configure/helper/repoAnalysisHelper.ts
```diff
@@ -46,32 +47,35 @@ export class RepoAnalysisHelper {
                 let applicationSettings: RepositoryAnalysisApplicationSettings = new RepositoryAnalysisApplicationSettings();
                 applicationSettings.language = analysis.language;
 
-                if(!!analysis.settings && !!analysis.buildTargetName) {
```
bishal-pdMSFT: rename `analysis` to `applicationSetting`

src/configure/helper/repoAnalysisHelper.ts
```diff
@@ -46,32 +47,35 @@ export class RepoAnalysisHelper {
                 let applicationSettings: RepositoryAnalysisApplicationSettings = new RepositoryAnalysisApplicationSettings();
```
bishal-pdMSFT: `RepositoryAnalysisApplicationSettings` should be renamed to just `ApplicationSettings` as this a VS Code model and nothing specific to do with repo analysis

src/configure/templates/githubWorkflowTemplates/nodejsWithGruntOnLinuxWebApp.yml
```diff
@@ -27,7 +27,7 @@ jobs:
 
     # install dependencies, build, and test
     - name: npm install, build
-      working-directory: {{{ repositoryAnalysisApplicationSettings.settings.nodePackageFileDirectory }}}
+      working-directory: {{{ workingDirectory }}}
       run: |
         npm install
         grunt --gruntfile gruntfile.js
```
bishal-pdMSFT: this should be updated to point to applicationSettings setting

bishal-pdMSFT: similarly for others as well

src/configure/helper/repoAnalysisHelper.ts
```diff
@@ -46,32 +47,35 @@ export class RepoAnalysisHelper {
...
-                            applicationSettings.settings.nodeGulpFilePath = analysis.settings[RepoAnalysisConstants.GulpFilePath];
+                if(!!analysis.settings){
+                    applicationSettings.settings.workingDirectory = !!analysis.settings.workingDirectory ? analysis.settings.workingDirectory : workspacePath;
+                    if(!!analysis.buildTargetName) {
+                        applicationSettings.buildTargetName = analysis.buildTargetName;
+                        if (analysis.language === SupportedLanguage.NODE) {
+                            applicationSettings.settings.nodePackageFilePath = analysis.settings[RepoAnalysisConstants.PackageFilePath];
+                            applicationSettings.settings.nodePackageFileDirectory = path.dirname(analysis.settings[RepoAnalysisConstants.PackageFilePath]);
+                            if (analysis.buildTargetName === RepoAnalysisConstants.Gulp && !!analysis.settings[RepoAnalysisConstants.GulpFilePath]) {
+                                applicationSettings.settings.nodeGulpFilePath = analysis.settings[RepoAnalysisConstants.GulpFilePath];
```
bishal-pdMSFT: where are these being used? Shouldn't they be used while creating workflow? I have left a comment there as well.

src/configure/helper/repoAnalysisHelper.ts
```diff
@@ -46,32 +47,35 @@ export class RepoAnalysisHelper {
...
 
-                if(!!analysis.settings && !!analysis.buildTargetName) {
-                    applicationSettings.buildTargetName = analysis.buildTargetName;
-                    if (analysis.language === SupportedLanguage.NODE) {
-                        applicationSettings.settings.nodePackageFilePath = analysis.settings[RepoAnalysisConstants.PackageFilePath];
-                        applicationSettings.settings.nodePackageFileDirectory = path.dirname(analysis.settings[RepoAnalysisConstants.PackageFilePath]);
-                        if (analysis.buildTargetName === RepoAnalysisConstants.Gulp && !!analysis.settings[RepoAnalysisConstants.GulpFilePath]) {
-                            applicationSettings.settings.nodeGulpFilePath = analysis.settings[RepoAnalysisConstants.GulpFilePath];
+                if(!!analysis.settings){
+                    applicationSettings.settings.workingDirectory = !!analysis.settings.workingDirectory ? analysis.settings.workingDirectory : workspacePath;
```
bishal-pdMSFT: Why do we need this variable called `applicationSettings`. can't we directly use `analysis` object?

bishal-pdMSFT:
@issacnitin can you please review this PR as well. Especially review the workflow file changes.

src/configure/configure.ts
bishal-pdMSFT: based on the logic of our repo analysis, this filter can yield more than one elements - think of case where there are two gruntfiles under same working directory (package.json). In such cases there will be elements in repoAnalysisResult which will match the filter condition.
We will have to drill down further and ask user to select specific gulp/grunt or other such file to zero down on one element

src/configure/configure.ts
vinodkumar3: We don't want same entry more than once in our array. Hence we are first converting string to HashSet and then converting it back to array.

src/configure/configure.ts
vinodkumar3: No, that is why I've used HashSet above to fetch unique list of working directories

src/configure/helper/repoAnalysisHelper.ts
vinodkumar3: Analysis object can be changed if service chose to update their model but applicationSetting will be same which is known by rest of code in our extension.

src/configure/helper/repoAnalysisHelper.ts
vinodkumar3: GitHub Templates

src/configure/templates/githubWorkflowTemplates/nodejsWithGruntOnLinuxWebApp.yml
vinodkumar3: No, working directory is not a part of applicationSettings since we are setting it as a property of pipeline configuration.

src/configure/helper/repoAnalysisHelper.ts
vinodkumar3: we are constructing applicationSettings already, this will induce confusion.

src/configure/configure.ts
vinodkumar3: We are setting repositoryAnalysisApplicationSettings and workingDirectory both

src/configure/configure.ts
vinodkumar3: Done

src/configure/configure.ts
vinodkumar3: Added a fail fast check

package.json
vinodkumar3: Sure

PR #66, opened 2020-03-23
Create ci-scan.yml

vineetmimrot:


.azure-pipelines/ci-scan.yml
```diff
@@ -0,0 +1,25 @@
+pool:
```
bishal-pdMSFT: how will this yml get consumed? I don't see it being consumed in azure-pipelines.yml?

.azure-pipelines/ci-scan.yml
```diff
@@ -0,0 +1,25 @@
...
+  name: Hosted VS2017
```
bishal-pdMSFT: can it run on hosted ubuntu? build/package pipelines runs on ubuntu

vineetmimrot: I don't think that will matter, as credscan & TSLint just scans the code and doesn't build it.

.azure-pipelines/ci-scan.yml
vineetmimrot: There will be separate ci for this
https://dev.azure.com/mseng/AzureDevOps/_build?definitionId=10036&_a=summary

PR #68, opened 2020-03-27
Adding telemetry

kanika1894:


src/configure/configurers/azurePipelineConfigurer.ts
```diff
@@ -230,7 +230,8 @@ export class AzurePipelineConfigurer implements Configurer {
     public async getPathToManifestFile(inputs: WizardInputs, localGitRepoHelper: LocalGitRepoHelper, fileName: string): Promise<string> { return null; }
 
     public async checkInPipelineFilesToRepository(filesToCommit: string[], inputs: WizardInputs, localGitRepoHelper: LocalGitRepoHelper): Promise<string> {
-
+        
+        telemetryHelper.setTelemetry(TelemetryKeys.SelectedCICDProvider, constants.azurePipeline);
```
bishal-pdMSFT: I think this is quite late in the flow. Although technically right to do here, it would make more sense to set telemetry at the place where we decide which pipeline type to use. And that place, I believe, is [here](https://github.com/microsoft/vscode-deploy-azure/blob/667891ccce189b914912b9b8a9181e3111726820/src/configure/configure.ts#L91). The pipelineConfigurer is different for AzP and GhW

kanika1894: updated

bishal-pdMSFT:
If you want to ship this telemetry change, then you also need to bump up the extension version

PR #72, opened 2020-04-06
Added tslint configuration for local VSCode

vineetmimrot:
The user needs to do run `npm install` in the root directory and restart VSCode and after that TSLint will start running automatically.


tslint.json
```diff
@@ -10,7 +10,45 @@
...
+			  "import-sources-order": "case-insensitive",
+			  "named-imports-order": "case-insensitive"
+			}
+		],
+		"no-constant-condition": [
+			true,
+			{
+				"checkLoops": false
+			}
+		]
```
vineetmimrot, line 50: I have suppressed many TSLint errors via turning off these rules as these are very basic sanity things and not needed at all. After fixing all the current TSLint errors, we can enable this.

tslint.json
```diff
@@ -10,7 +10,45 @@
...
 	},
-	"defaultSeverity": "warning"
+	"defaultSeverity": "warning",
+	"extends": ["tslint-microsoft-contrib/recommended", "tslint:recommended"]
```
vineetmimrot, line 53: `tslint-microsoft-contrib/recommended` show errors corresponding to ci build scan i.e. sdl required.

`tslint:recommended` show general TSLint errors.

package.json
```diff
@@ -90,18 +90,20 @@
...
-        "@types/js-yaml": "^3.12.1",
         "ajv": "^6.9.1",
         "assert": "1.4.1",
         "mocha": "5.2.0",
-        "tslint": "5.8.0",
         "ts-node": "7.0.1",
+        "tslint": "5.8.0",
+        "tslint-microsoft-contrib": "^6.2.0",
         "typescript": "3.3.1",
+        "typescript-tslint-plugin": "^0.5.5",
```
bishal-pdMSFT, line 106: for my understanding - what is difference between tslint and typescript-tslint-plugin?

bishal-pdMSFT: and where is this plugin used?

tslint.json
bishal-pdMSFT: @vineetmimrot how will we track fixing these warning?

tslint.json
bishal-pdMSFT, line 53: if we specify "tslint-microsoft-contrib/recommended" in extends here, 'do we also need to specify rulesDirectory in line no 2? what is the different between the two?

tslint.json
```diff
@@ -10,7 +10,45 @@
...
+		"no-string-based-set-immediate": true,
+		"no-string-based-set-interval": true,
+		"no-string-based-set-timeout": true,
+		"no-eval": true,
+		"quotemark": false,
+		"no-relative-imports": false,
+		"max-line-length": [
+			true,
+			{
+			  "limit": 200,
```
bishal-pdMSFT, line 32: I would prefer keeping a smaller max length

package.json
vineetmimrot: It is VSCode plugin, which runs the TSLint in the IDE.

tslint.json
vineetmimrot: Line no 2, specifies the rules directory, but it doesn't say which rules to use from the directory.
Extends specifies which rule config files to use and tool will search into rules directory.

tslint.json
vineetmimrot: by default it is 120, I made it 200. 
Will change to 135

PR #80, opened 2020-04-15
Mandatory Authorized calls for Repo Analysis and Service URL fetch for backward compat

issacnitin:


src/configure/configure.ts
```diff
@@ -154,6 +157,23 @@ class Orchestrator {
         }
     }
 
+    private async getGithubPatToken(): Promise<void> {
+        const redirectHelper = new RedirectLinkHelper();
+        await redirectHelper.loadAll();
+        if (this.inputs.sourceRepository.repositoryProvider === "github" || redirectHelper.repoAnalysisServiceFramework === ServiceFramework.Moda) {
```
vineetmimrot, line 163: Should we throw an error in else case as it will be undesired behaviour?

src/configure/helper/redirectLinkHelper.ts
```diff
@@ -0,0 +1,38 @@
+import { RestClient } from "typed-rest-client";
+
+export enum ServiceFramework {
+    Moda,
+    Vssf
+}
+
+export class RedirectLinkHelper {
+
+    public repoAnalysisRedirectUrl: string  = "https://aka.ms/AA87j8k";
```
bishal-pdMSFT, line 10: should we not use a fwd link instead of aka.ms?

src/configure/helper/redirectLinkHelper.ts
```diff
@@ -0,0 +1,38 @@
...
+            if ((response.message.statusCode === 301 || response.message.statusCode === 302)){
+                this.repoAnalysisUrl = response.message.headers["location"];
+            } else {
+                throw Error("Invalid response from url " + this.repoAnalysisRedirectUrl);
+            }
+            if (!this.repoAnalysisUrl.includes("portalext.visualstudio.com")) {
+                this.repoAnalysisServiceFramework = ServiceFramework.Moda;
+            }
+        } catch (err) {
+            // Add telemetry
```
bishal-pdMSFT, line 35: please add telemetry. don't leave it as todo.

src/configure/helper/redirectLinkHelper.ts
```diff
@@ -0,0 +1,38 @@
...
+            const response = await restClient.client.get(this.repoAnalysisRedirectUrl, requestOptions);
+            if ((response.message.statusCode === 301 || response.message.statusCode === 302)){
+                this.repoAnalysisUrl = response.message.headers["location"];
+            } else {
+                throw Error("Invalid response from url " + this.repoAnalysisRedirectUrl);
+            }
+            if (!this.repoAnalysisUrl.includes("portalext.visualstudio.com")) {
+                this.repoAnalysisServiceFramework = ServiceFramework.Moda;
+            }
+        } catch (err) {
```
bishal-pdMSFT, line 34: in case of an exception, use vssf service url as default

src/configure/helper/repoAnalysisHelper.ts
```diff
@@ -1,24 +1,28 @@
...
+import { ModaClient } from '../clients/modaClient';
 import { PortalExtensionClient } from "../clients/portalExtensionClient";
 import { AzureSession, GitRepositoryParameters, RepositoryAnalysisApplicationSettings, RepositoryAnalysisParameters, RepositoryAnalysisRequest, RepositoryDetails, RepositoryProvider, SupportedLanguage } from "../model/models";
 import { RepoAnalysisConstants } from "../resources/constants";
+import { RedirectLinkHelper, ServiceFramework } from './redirectLinkHelper';
+
 
 export class RepoAnalysisHelper {
     private portalExtensionClient: PortalExtensionClient;
+    private modaClient: ModaClient;
```
bishal-pdMSFT, line 11: maintaining two different clients is problematic. We should define an interface for the repositoryAnalysisClient and then have two implementations of that - PortalExtensionRespositoryAnalysisClient and ModaRepositoryAnalysisClient. Once such a client is created, the interface will help in using it anywhere in code, otherwise everywhere we have to pass both portalExtensionClient and modaClient

src/configure/helper/repoAnalysisHelper.ts
```diff
@@ -1,24 +1,28 @@
...
+    private redirectHelper: RedirectLinkHelper;
+    private githubPatToken?: string;
 
-    constructor(azureSession: AzureSession) {
+    constructor(azureSession: AzureSession, githubPatToken?: string) {
         this.portalExtensionClient = new PortalExtensionClient(azureSession.credentials);
```
bishal-pdMSFT, line 16: why create it here? it should be created if the vssf url is being used.

src/configure/helper/redirectLinkHelper.ts
```diff
@@ -0,0 +1,38 @@
+import { RestClient } from "typed-rest-client";
+
+export enum ServiceFramework {
+    Moda,
+    Vssf
+}
+
+export class RedirectLinkHelper {
```
bishal-pdMSFT, line 8: this should be called `RemoteServiceUrlHelper` so as not to tie it with underlying logic

src/configure/helper/redirectLinkHelper.ts
```diff
@@ -0,0 +1,38 @@
...
+
+    public repoAnalysisRedirectUrl: string  = "https://aka.ms/AA87j8k";
+    public repoAnalysisUrl: string = "";
```
bishal-pdMSFT, line 11: IMO it looks unnecessary to have these as public properties. These should be private and can be used as a cache (so that mulitple invocation of loadAll() does not make redirect calls again and again). The load.. functions should return a data type containing url and framework and the caller should use the return value directly instead of depending on these public properties.

bishal-pdMSFT: Infact it would have been okay to not store anything in this class and use this purely as helper to run some logic. If in future we need to make multiple invocations, then we would think about cache etc. otherwise we are unnecessarily storing things.

src/configure/configure.ts
```diff
@@ -154,6 +157,23 @@ class Orchestrator {
         }
     }
 
+    private async getGithubPatToken(): Promise<void> {
+        const redirectHelper = new RedirectLinkHelper();
+        await redirectHelper.loadAll();
```
bishal-pdMSFT, line 162: you are calling this at multiple places -  then loadAll seems like a wrong pattern. LoadAll signifies that it should be loaded just once. Either load it really just once or directly expose a function called getRepositoryAnalysisUrl and implement a caching as suggested in other comment

src/configure/configurers/azurePipelineConfigurer.ts
```diff
@@ -42,16 +42,6 @@ export class AzurePipelineConfigurer implements Configurer {
 
     public async getInputs(inputs: WizardInputs): Promise<void> {
         try {
-            if (inputs.sourceRepository.repositoryProvider === RepositoryProvider.Github) {
```
bishal-pdMSFT, line 45: we should not make any change in AzurePipelines creation flow and this guy will not need PAT for repo analysis.

src/configure/configurers/azurePipelineConfigurer.ts
```diff
@@ -42,16 +42,6 @@ export class AzurePipelineConfigurer implements Configurer {
...
-                inputs.githubPATToken = await this.controlProvider.showInputBox(constants.GitHubPat, {
-                    placeHolder: Messages.enterGitHubPat,
-                    prompt: Messages.githubPatTokenHelpMessage,
```
bishal-pdMSFT, line 48: One more reason to not touch this code is that it uses a different prompt message than the github workflow flow one.

src/configure/configure.ts
```diff
@@ -154,6 +157,23 @@ class Orchestrator {
...
+        if (this.inputs.sourceRepository.repositoryProvider === "github" || redirectHelper.repoAnalysisServiceFramework === ServiceFramework.Moda) {
+            this.inputs.githubPATToken = await this.controlProvider.showInputBox(constants.GitHubPat, {
+                placeHolder: Messages.enterGitHubPat,
+                prompt: Messages.githubPatTokenHelpMessage,
```
bishal-pdMSFT, line 166: the prompt message is different in two cases as one requires additional scopes. please handle this.

src/configure/configure.ts
bishal-pdMSFT, line 163: Shouldn't we get pat irrespective of whether it is moda or vssf service? Are you getting pat token in case of vssf service somewhere else?

src/configure/configure.ts
```diff
@@ -154,6 +157,23 @@ class Orchestrator {
...
+                validateInput: (inputValue) => {
+                    return !inputValue ? Messages.githubPatTokenErrorMessage : null;
+                }
+            });
+            if (!!this.inputs.githubPATToken) {
+                this.inputs.githubPATToken = this.inputs.githubPATToken.toLowerCase();
```
bishal-pdMSFT, line 172: why the case change? we should trust user to do right thing. On the other side it can break the PAT if github decides to issue pat with capital case letters.

src/configure/clients/modaClient.ts
```diff
@@ -0,0 +1,26 @@
...
+import { RepositoryAnalysisRequest } from "../model/models";
+
+export class ModaClient{
+    private restClient: RestClient;
+    private githubPat: string;
+    private pathUrl: string;
+
+    constructor(url: string, githubPat: string) {
+        const u = vscodeUri.parse(url);
+        this.restClient = new RestClient("deploy-to-azure", u.scheme + "://" + u.authority);
```
bishal-pdMSFT, line 12: what is the second parameter? is it creating a value of type `ServiceClientCredentials`? if so, why not take this type in the constructor. this way both PEClient and ModaClient will have same constructor signature- - and easier to derive from same interface.

src/configure/helper/repoAnalysisHelper.ts
```diff
@@ -28,7 +32,15 @@ export class RepoAnalysisHelper {
             repositoryAnalysisRequestBody.Repository = repositoryDetails;
             repositoryAnalysisRequestBody.WorkingDirectory = workspacePath;
 
-            repositoryAnalysisResponse = await this.portalExtensionClient.getRepositoryAnalysis(repositoryAnalysisRequestBody);
+            if (this.redirectHelper.repoAnalysisServiceFramework === ServiceFramework.Vssf) {
+                repositoryAnalysisRequestBody.Repository.authorizationInfo.scheme = "Token";
```
bishal-pdMSFT, line 36: Wouldn't we need these two lines for moda client call as well?

bishal-pdMSFT: otherwise you are not passing token for service to make authorized calls to github repo

src/configure/configure.ts
issacnitin: It is a valid flow - Azure Repo + Azure Pipeline. In that scenario we don't call Repo Analysis

src/configure/clients/modaClient.ts
issacnitin: no it's not creating value of type ServiceClientCredentials. Second parameter is the base url, in our case "https://aka.ms"

src/configure/configure.ts
issacnitin: In case of vssf service, we need to get PAT only for the flows when repository is of type github. For Azure Repo  + Azure Pipelines flow, we don't need

src/configure/configure.ts
issacnitin: Github right now doesn't support PAT with capital letters. So this is required. Agreed, it might break if Github changes to capital letters, but that's very unlikely as a lot of systems depending on GitHub pat will break

src/configure/helper/redirectLinkHelper.ts
issacnitin: Yes. If we need to cache we can implement this as a singleton.

src/configure/helper/repoAnalysisHelper.ts
issacnitin: In case of Moda we're authenticating using github pat, so this can be avoided as it's redundant

src/configure/configure.ts
bishal-pdMSFT: But why take this extra burden which, even though unlikely, has a chance to break.
If user is inputting a wrong cased PAT he will end up in failure. Lets not try to fix user's error.

src/configure/helper/repoAnalysisHelper.ts
bishal-pdMSFT: this is not about authneticating to moda based service. This is about template service using eight credential when making calls to github repo. For that it needs a credential parameter along with repo information.

src/configure/configure.ts
issacnitin: The PAT token can by default be used to authenticate to github API. Additional scopes needed are repo read access for github workflow, which is already being informed to the user

src/configure/helper/redirectLinkHelper.ts
issacnitin: I couldn't find any reason why fwd link could be better, shall we sent out a mail to the owner team for their opinion? It'll help us understand if there are any security/compliance issues as well in using their product for our purpose.

src/configure/helper/repoAnalysisHelper.ts
issacnitin: Yes I was thinking of using token from auth header, why'd you feel so? Are we changing it in template service as well?

src/configure/configure.ts
issacnitin: Alright

src/configure/helper/repoAnalysisHelper.ts
issacnitin: I mean, ideally calling moda service is authenticating to github, so we can use the same auth header for all internal calls as well

src/configure/clients/modaRepositoryAnalysisClient.ts
```diff
@@ -0,0 +1,27 @@
...
+    private pathUrl: string;
+
+    constructor(url: string, githubPat: string) {
+        const u = vscodeUri.parse(url);
+        this.restClient = new RestClient("deploy-to-azure", u.scheme + "://" + u.authority);
+        this.pathUrl = u.path;
+        this.githubPat = githubPat;
+    }
+
+    public async getRepositoryAnalysis(body: RepositoryAnalysisRequest): Promise<IRestResponse<unknown>> {
```
bishal-pdMSFT, line 18: why `unknown` type. I guess we have a well defined type for repo analysis result.

src/configure/clients/modaRepositoryAnalysisClient.ts
```diff
@@ -0,0 +1,27 @@
...
+        const requestOptions = {
+            acceptHeader: "application/json",
+            additionalHeaders: {
+                "Authorization": "Bearer " + this.githubPat
+            }
+        };
+        return this.restClient.create(this.pathUrl, body, requestOptions);
```
bishal-pdMSFT, line 25: `create` or `sendRequest` ?

src/configure/clients/repositoryAnalyisClient.ts
```diff
@@ -0,0 +1,5 @@
+import { RepositoryAnalysisRequest } from "../model/models";
+
+export interface IRepositoryAnalysisClient {
+    getRepositoryAnalysis(body: RepositoryAnalysisRequest): Promise<any>;
```
bishal-pdMSFT, line 4: return type should not be `any` if possible.

src/configure/configure.ts
```diff
@@ -12,7 +12,9 @@ import { ControlProvider } from './helper/controlProvider';
 import { AzureDevOpsHelper } from './helper/devOps/azureDevOpsHelper';
 import { GitHubProvider } from './helper/gitHubHelper';
 import { LocalGitRepoHelper } from './helper/LocalGitRepoHelper';
+import { RemoteServiceUrlHelper, ServiceFramework } from './helper/remoteServiceUrlHelper';
 import { RepoAnalysisHelper } from './helper/repoAnalysisHelper';
+//import { RepoAnalysisHelper } from './helper/repoAnalysisHelper';
```
bishal-pdMSFT, line 17: remove this

src/configure/clients/modaRepositoryAnalysisClient.ts
issacnitin: create is for POST request

src/configure/configure.ts
bishal-pdMSFT: Don't we need additional scope for pushing workflow file in case of github workflow?

src/configure/configure.ts
```diff
@@ -427,11 +442,13 @@ class Orchestrator {
...
-        var repoAnalysisResult = await repoAnalysisHelper.getRepositoryAnalysis(this.inputs.sourceRepository,
-            this.inputs.pipelineConfiguration.workingDirectory.split('/').join('\\'));
+        const repoAnalysisHelper = new RepoAnalysisHelper(this.inputs.azureSession, this.inputs.githubPATToken);
+        let repoAnalysisResult = null;
+        if (this.inputs.sourceRepository.repositoryProvider === RepositoryProvider.Github) {
+            repoAnalysisResult = await repoAnalysisHelper.getRepositoryAnalysis(this.inputs.sourceRepository, this.inputs.pipelineConfiguration.workingDirectory.split('\\').join('/'));
+        }
 
-        extensionVariables.templateServiceEnabled = true;
+        extensionVariables.templateServiceEnabled = false;
```
bishal-pdMSFT, line 451: Whoops! we could have shipped this by mistake! Thanks for catching this.

src/configure/configure.ts
```diff
@@ -427,11 +442,13 @@ class Orchestrator {
     }
 
     private async getSelectedPipeline(): Promise<void> {
-        var repoAnalysisHelper = new RepoAnalysisHelper(this.inputs.azureSession);
-        var repoAnalysisResult = await repoAnalysisHelper.getRepositoryAnalysis(this.inputs.sourceRepository,
-            this.inputs.pipelineConfiguration.workingDirectory.split('/').join('\\'));
+        const repoAnalysisHelper = new RepoAnalysisHelper(this.inputs.azureSession, this.inputs.githubPATToken);
+        let repoAnalysisResult = null;
+        if (this.inputs.sourceRepository.repositoryProvider === RepositoryProvider.Github) {
+            repoAnalysisResult = await repoAnalysisHelper.getRepositoryAnalysis(this.inputs.sourceRepository, this.inputs.pipelineConfiguration.workingDirectory.split('\\').join('/'));
```
bishal-pdMSFT, line 448: Are we making sure that we rely on `repoAnalysisResult` only in case of github repo.

src/configure/helper/repoAnalysisHelper.ts
```diff
@@ -1,34 +1,45 @@
...
             repositoryAnalysisRequestBody.WorkingDirectory = workspacePath;
-
-            repositoryAnalysisResponse = await this.portalExtensionClient.getRepositoryAnalysis(repositoryAnalysisRequestBody);
+            repositoryAnalysisRequestBody.Repository.authorizationInfo = {
+                scheme: "Token",
+                parameters: {
+                    accesstoken: this.githubPatToken
+                }
+            };
+            if (serviceDefinition.serviceFramework === ServiceFramework.Vssf) {
```
bishal-pdMSFT, line 38: this if condition is not required anymore. that is the beauty of using an abstract interface :)

src/configure/helper/repoAnalysisHelper.ts
```diff
@@ -90,4 +101,14 @@ export class RepoAnalysisHelper {
     private GetRelativePath(workingDirectory: string, filePath: string): string{
         return path.relative(workingDirectory, filePath).split(path.sep).join('/');
     }
+
+    private getClient(serviceDefinition: IServiceUrlDefinition) {
```
bishal-pdMSFT, line 105: please specify return type

src/configure/helper/repoAnalysisHelper.ts
```diff
@@ -90,4 +101,14 @@ export class RepoAnalysisHelper {
...
+        let client = null;
+        if (serviceDefinition.serviceFramework == ServiceFramework.Vssf) {
+            client = new PortalExtensionRepositoryAnalysisClient(this.azureSession.credentials);
```
bishal-pdMSFT, line 108: We should pass the PE repo analysis url as input here as well. And remove the url from L16 in PortalExtensionRepositoryAnalysisClient.ts as it is duplicate of the PR url already specified in url helper class.

src/configure/configure.ts
issacnitin: Yes that's what this check is for

src/configure/helper/repoAnalysisHelper.ts
issacnitin: Thanks!

src/configure/configure.ts
issacnitin: No, read/write access is what is needed for secrets api, pushing workflow file etc.

PR #83, opened 2020-04-21
Right click scenario

kanika1894:


src/configure/templateInputHelper/InputControl.ts
```diff
@@ -103,6 +104,21 @@ export class InputControl {
         }
     }
 
+    public getPropertyValue(propertyName: string, inputs?: StringMap<any>): string {
+        var properties = this.inputDescriptor.properties;
+        var value: string;
+
+        if (!!properties) {
+            value = properties[propertyName];
+            if (!!value && !!inputs && typeof value === "string") {
```
anuragc617, line 113: change the condition to:
`if( !!value && !!inputs ) {
  return MustacheHelper.render(value, { inputs: inputs }); 
}`

and remove else part.

PR #84, opened 2020-04-23
Code refactoring

kanika1894:
This will now support both local and remote template flow.
The flag : templateServiceEnabled is used for the same.
Also, as of now, random weights are added to the local templates, which can be changed later,


anuragc617:
Can you update your branch and fix the build issues.

src/configure/helper/templateHelper.ts
```diff
@@ -112,35 +139,50 @@ export async function analyzeRepoAndListAppropriatePipeline(repoPath: string, re
...
+    var localPipelineTemplatesPromise: Promise<LocalPipelineTemplate> = await this.analyzeRepoAndListAppropriatePipeline(repoPath, repositoryProvider, repoAnalysisParameters);
 
     let serviceClient = new TemplateServiceClient(azureSession.credentials);
-    templateResult = await serviceClient.getTemplates(repoAnalysisParameters);
-    templateResult = templateResult.sort((a, b) => {
-        if (a.templateWeight > b.templateWeight) { return 1; }
-        else { return -1; }
-    });
-    return templateResult;
+    var templateResultPromise = await serviceClient.getTemplates(repoAnalysisParameters);
```
anuragc617, line 151: getTemplates from TemplateService only if repoAnalysisParameters is not null and contains repositoryAnalysisApplicationSettingsList

src/configure/helper/templateHelper.ts
```diff
@@ -112,35 +139,50 @@ export async function analyzeRepoAndListAppropriatePipeline(repoPath: string, re
...
+
+    return Q.all([localPipelineTemplatesPromise, templateResultPromise])
```
anuragc617, line 153: We should add fallback logic to return local templates if we fail to get remote templates.

src/configure/configure.ts
```diff
@@ -408,98 +446,75 @@ class Orchestrator {
...
                     this.inputs.sourceRepository.repositoryProvider,
                     repoAnalysisResult,
                     this.inputs.pipelineConfiguration.params[constants.TargetResource])
             );
-        }
-        else {
-            appropriatePipelines = await vscode.window.withProgress(
+            appropriatePipelines = localPipelines;
+        } else {
+            repoAnalysisResult = {
```
bishal-pdMSFT, line 469: Why use a dummy result here? We now have repo analysis wired up. no?

src/configure/helper/templateHelper.ts
```diff
@@ -112,35 +139,50 @@ export async function analyzeRepoAndListAppropriatePipeline(repoPath: string, re
...
+                    workingDirectory: template.workingDirectory
+                };
+                pipelineTemplates.push(remoteTemplate);
+            });
+            // sorted by weight
+            pipelineTemplates = pipelineTemplates.sort((a, b) => {
+                if (a.templateWeight < b.templateWeight) { return 1; }
+                else { return -1; }
+            });
+            return pipelineTemplates.concat(localPipelineTemplates);
```
bishal-pdMSFT, line 173: Should we not filter out local templates if they  have same name?

src/configure/helper/templateHelper.ts
bishal-pdMSFT: +1

src/configure/configure.ts
```diff
@@ -408,98 +446,75 @@ class Orchestrator {
...
         extensionVariables.templateServiceEnabled = false;
+        var appropriatePipelines: PipelineTemplate[] = [];
 
-        let appropriatePipelines;
-        // TO:DO- Get applicable pipelines for the repo type and azure target type if target already selected
-
-        if (extensionVariables.templateServiceEnabled) {
-            repoAnalysisResult = null;
-            appropriatePipelines = await vscode.window.withProgress(
+        if (!extensionVariables.templateServiceEnabled) {
```
bishal-pdMSFT, line 458: is this condition sufficient? In case of Azure repos, we would still want to use local repo analysis and local templates.

bishal-pdMSFT: or may be this is right. So if I understand right, analyzeRepoAndListAppropriatePipeline2() merges both local template and remote templates, and hence for azure repos, it still will be able to list local azure pipeline templates. Similarly for AKS. Correct?

src/configure/configure.ts
```diff
@@ -408,98 +446,75 @@ class Orchestrator {
...
                     this.inputs.pipelineConfiguration.params[constants.TargetResource])
             );
+            appropriatePipelines = remotePipelines;
         }
 
+        let pipelineMap: Map<string, PipelineTemplate[]> = new Map();
+
+        appropriatePipelines.forEach(element => {
+            if (pipelineMap.has(element.label)) {
+                pipelineMap.get(element.label).push(element);
```
bishal-pdMSFT, line 495: Is the intention here is to merge local and remote templates? Two questions:
1. Why do we need to do that? If remote template with label exist, should we not ignore local template?
2. Even if we need this, should we not do this inside analyzeRepoAndListAppropriatePipeline2() as analyzeRepoAndListAppropriatePipeline does not return both type of templates?

src/configure/configure.ts
```diff
@@ -408,98 +446,75 @@ class Orchestrator {
...
         else {
-            if (extensionVariables.templateServiceEnabled) {
-                this.inputs.pipelineConfiguration.templateInfo = appropriatePipelines[0];
-            }
-            else {
-                this.inputs.pipelineConfiguration.template = appropriatePipelines[0];
-            }
+            this.inputs.potentialTemplates = pipelineMap.get(pipelineLabels[0]);
         }
+        this.inputs.pipelineConfiguration.template = this.inputs.potentialTemplates[0];
```
bishal-pdMSFT, line 517: Why should this be done here? Should it not be done when we are selecting template based on potential templates?

src/configure/configure.ts
```diff
@@ -511,13 +526,8 @@ class Orchestrator {
             //Get languageSettings (corresponding to language of selected settings) provided by RepoAnalysis
             await this.updateRepositoryAnalysisApplicationSettings(repoAnalysisResult);
         }
+        telemetryHelper.setTelemetry(TelemetryKeys.ChosenTemplate, this.inputs.pipelineConfiguration.template.label);
```
bishal-pdMSFT, line 529: this also

src/configure/configure.ts
```diff
@@ -124,46 +187,21 @@ class Orchestrator {
             await this.getAzureSession();
             await this.getGithubPatToken();
             await this.getSelectedPipeline();
+            if (!resourceNode) {
+                let selectedResource = await this.getAzureResource();
```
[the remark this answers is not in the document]
bishal-pdMSFT: I will Let Kanika reply here, but as per my understanding getAzureresource() internally takes care of that by throwing `Messages.setupAlreadyConfigured`

src/configure/configurers/AksAzureResourceSelector.ts
```diff
@@ -0,0 +1,12 @@
+import { GenericResource } from "azure-arm-resource/lib/resource/models";
+import { TargetResourceType } from "../model/models";
+import { IAzureResourceSelector } from "./IAzureResourceSelector";
+
+export class AksAzureResourceSelector implements IAzureResourceSelector {
+     async getAzureResource(inputs): Promise<GenericResource> {
+          return { type: TargetResourceType.AKS };
```
bishal-pdMSFT, line 7: I did not understand this. Should we not be showing list of AKS resources to user?

src/configure/configurers/IAzureResourceSelector.ts
```diff
@@ -0,0 +1,8 @@
+import { GenericResource } from "azure-arm-resource/lib/resource/models";
+
+export interface IAzureResourceSelector{
+
```
bishal-pdMSFT, line 4: remove unnecessary blank lines

src/configure/configurers/ResourceSelectorFactory.ts
```diff
@@ -0,0 +1,20 @@
+import { PipelineConfiguration, TargetResourceType } from "../model/models";
+import { AksAzureResourceSelector } from "./AksAzureResourceSelector";
+import { IAzureResourceSelector } from "./IAzureResourceSelector";
+import { WebAppAzureResourceSelector } from "./WebAppAzureResourceSelector";
+
+export class ResourceSelectorFactory {
+
+    public static getAzureResourceSelector(pipelineConfiguration: PipelineConfiguration): IAzureResourceSelector {
+        if (pipelineConfiguration.template.label === "Containerized application to AKS") {
```
bishal-pdMSFT, line 9: If I am not wrong, the template is not yet decided, then why depend on `pipelineConfiguration.template`? We should rather use `this.inputs.potentialTemplates`. No?

src/configure/configurers/AksAzureResourceSelector.ts
```diff
@@ -0,0 +1,12 @@
...
+
```
bishal-pdMSFT, line 8: Also, we are not checking whether a pipeline is already configured for a AKS resource (namespace?). Is that because we can always create a new pipeline targeting a new namespace? If yes, it sounds okay. But then do we never show browse experience for AKS ? @anuragc617 your thoughts

bishal-pdMSFT: BTW I am digressing here so this is a low priority comment.

src/configure/model/models.ts
```diff
@@ -35,6 +35,7 @@ export class WizardInputs {
     azureSession: AzureSession;
     subscriptionId: string;
     githubPATToken?: string;
+    potentialTemplates?: PipelineTemplate[];
```
bishal-pdMSFT, line 38: Why does it need to be a list of templates? It should just a `templateLabel`. And we should maintain a seprate templateMapper which maps a label to _n_ templates

src/configure/configure.ts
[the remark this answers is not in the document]
kanika1894: That's handled in WebAppResourceSelector.ts
```
     if (await appServiceClient.isScmTypeSet((<GenericResource>selectedResource.data).id)) {
               await openBrowseExperience((<GenericResource>selectedResource.data).id);
               throw new Error(Messages.setupAlreadyConfigured);
          }
```

src/configure/configure.ts
```diff
@@ -116,6 +116,69 @@ class Orchestrator {
...
+                if (shortlistedTemplates.length > 1) {
+                    inputs.pipelineConfiguration.template = shortlistedTemplates.find((template) => template.templateType === TemplateType.remote);
+                }
+                else {
+                    inputs.pipelineConfiguration.template = shortlistedTemplates[0];
+                }
+                break;
+
+            default:
+                let appServiceClient = new AppServiceClient(this.inputs.azureSession.credentials, this.inputs.azureSession.environment, this.inputs.azureSession.tenantId, this.inputs.subscriptionId);
```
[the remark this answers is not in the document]
kanika1894: removed

src/configure/configure.ts
```diff
@@ -124,46 +187,21 @@ class Orchestrator {
             await this.getAzureSession();
             await this.getGithubPatToken();
             await this.getSelectedPipeline();
+            if (!resourceNode) {
+                let selectedResource = await this.getAzureResource();
+                await this.selectTemplate(selectedResource, this.inputs);
```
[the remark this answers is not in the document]
kanika1894: updated

src/configure/configure.ts
```diff
@@ -116,6 +116,69 @@ class Orchestrator {
...
+                this.continueOrchestration = false;
+                return;
+            }
+            else {
+                throw err;
+            }
+        }
+    }
+
+    private async selectTemplate(resource: GenericResource, inputs: WizardInputs) {
```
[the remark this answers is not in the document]
kanika1894: But as we discussed, for AKS, a lot of things are done in the one function itself which is being called here

src/configure/model/templateModels.ts
```diff
@@ -1,26 +1,44 @@
...
+
+export interface PipelineTemplateMetadata {
+    templateId: string;
+    workingDirectory: string;
+    templateWeight: number;
+    templateDescription: string;
+    attributes: TemplateAttributes;
+}
+
+export interface RemotePipelineTemplate extends PipelineTemplate {
```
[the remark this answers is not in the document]
kanika1894: won't we able to get it from PipelineTemplate?
will replicating it in RemoteTemplate be the right way?

src/configure/model/templateModels.ts
```diff
@@ -1,26 +1,44 @@
-import { ExtendedPipelineTemplate } from "./Contracts";
 import { AzureConnectionType, ServiceConnectionType, TargetKind, TargetResourceType } from "./models";
 
+export enum TemplateType {
+    remote,
```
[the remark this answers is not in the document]
kanika1894: done

src/configure/helper/templateHelper.ts
```diff
@@ -41,11 +41,36 @@ export async function mergingRepoAnalysisResults(repoPath: string, repositoryPro
...
+
+export function getTargetType(template: PipelineTemplateMetadata) {
+    if (template.attributes.deployTarget.toLowerCase().includes("webapp")) {
+        return TargetResourceType.WebApp;
+    }
+    return TargetResourceType.None;
+}
+
+export function getTargetKind(template: PipelineTemplateMetadata) {
+
```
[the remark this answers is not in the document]
kanika1894: done

src/configure/helper/templateHelper.ts
```diff
@@ -41,11 +41,36 @@ export async function mergingRepoAnalysisResults(repoPath: string, repositoryPro
...
+    var targetKind: TargetKind;
+    if (template.attributes.deployTarget.toLowerCase().includes("container")) {
```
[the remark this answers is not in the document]
kanika1894: updated

src/configure/helper/templateHelper.ts
```diff
@@ -41,11 +41,36 @@ export async function mergingRepoAnalysisResults(repoPath: string, repositoryPro
     }
     return analysisResult;
 }
-export async function analyzeRepoAndListAppropriatePipeline(repoPath: string, repositoryProvider: RepositoryProvider, repoAnalysisParameters: RepositoryAnalysisParameters, targetResource?: GenericResource): Promise<PipelineTemplate[]> {
+
+export function getTargetType(template: PipelineTemplateMetadata) {
+    if (template.attributes.deployTarget.toLowerCase().includes("webapp")) {
```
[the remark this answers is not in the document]
kanika1894: updated

src/configure/helper/templateHelper.ts
kanika1894: updated

src/configure/helper/templateHelper.ts
kanika1894: Should I go back to the previous design where we will have a call made for local templates, and the condition of templateServiceEnabled will be used to determine whether we need to have a call for remote templates?

src/configure/helper/templateHelper.ts
kanika1894: At this stage, we aren't sure whether we'll proceed with local or remote template, as the user hasn't selected the webApp yet. So we are storing the potential templates as :

```
this.inputs.potentialTemplates = pipelineMap.get(selectedOption.label);
```

Once the user selects the webApp, will filter out the tenplates:

```
shortlistedTemplates = this.inputs.potentialTemplates.filter((template) => template.targetKind === resource.kind);
                if (shortlistedTemplates.length > 1) {
                    this.inputs.pipelineConfiguration.template = shortlistedTemplates.find((template) => template.templateType === TemplateType.REMOTE);
                }
```

src/configure/configure.ts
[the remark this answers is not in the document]
anuragc617: you are catching the exception in this.getAzureResource() and setting `this.continueOrchestration = false` but this variable is not checked here so it will continue to call selectTemplate() and then getAllPipelineTemplateInputs()

src/configure/configure.ts
kanika1894: The intention here is to have a list mapped by label, to have a dropdown with unique labels, in which the user will choose the template.
The local and remote filtering is done at a later stage, where the resource is selected.

src/configure/configure.ts
kanika1894: At this place we just actually get to know what will be the label of the template. The local/remote decision hasn't been taken yet.
But since we have to make a decision of AKS or webApp depending upon the template's label, that is why done this.

src/configure/configure.ts
kanika1894: The label gets decided and that's why we can store the telemetry as well.

src/configure/configurers/AksAzureResourceSelector.ts
anuragc617: @bishal-pdMSFT we don't show browse experience for AKS because we can always create a new pipeline even in the same namespace/ new namespace.

src/configure/model/templateModels.ts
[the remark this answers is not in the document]
anuragc617: We won't we able to get from PipelineTemplate. Yes replicate them to RemotTemplate, its right to have these properties there.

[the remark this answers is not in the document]
kanika1894: added

src/configure/helper/templateHelper.ts
```diff
@@ -112,35 +139,50 @@ export async function analyzeRepoAndListAppropriatePipeline(repoPath: string, re
...
-
 }
 
-export async function analyzeRepoAndListAppropriatePipeline2(azureSession: AzureSession, repoPath: string, repositoryProvider: RepositoryProvider, repoAnalysisParameters: RepositoryAnalysisParameters, targetResource?: GenericResource): Promise<PipelineTemplateMetadata[]> {
+export async function analyzeRepoAndListAppropriatePipeline2(azureSession: AzureSession, repoPath: string, repositoryProvider: RepositoryProvider, repoAnalysisParameters: RepositoryAnalysisParameters, targetResource?: GenericResource): Promise<PipelineTemplate[]> {
 
-    //TO:DO - Merge local repo analysis (Some changes in the definition of AnalysisResult required)
-    let templateResult: PipelineTemplateMetadata[] = [];
+    var pipelineTemplates: PipelineTemplate[] = [];
+    var localPipelineTemplatesPromise: Promise<LocalPipelineTemplate> = await this.analyzeRepoAndListAppropriatePipeline(repoPath, repositoryProvider, repoAnalysisParameters);
```
anuragc617, line 148: analyzeRepoAndListAppropriatePipeline returns Promise<LocalPipelineTemplate[]>, fix localPipelineTemplatesPromise type.

src/configure/helper/templateHelper.ts
anuragc617: @bishal-pdMSFT will templateService only return templates for github workflow? Are we planning to add AzurePipelines templates in template service?

src/configure/helper/templateHelper.ts
```diff
@@ -310,7 +338,9 @@ let azurePipelineTemplates: { [key in SupportedLanguage]: PipelineTemplate[] } =
                     "type": TemplateAssetType.AzureARMPublishProfileServiceConnection
                 }
             ],
-            azureConnectionType: AzureConnectionType.AzureRMPublishProfile
+            azureConnectionType: AzureConnectionType.AzureRMPublishProfile,
+            templateWeight: 123,
```
anuragc617, line 342: can you assign same weight for all localTemplates say -> 100. It will be easier in future if we need to change that.

src/configure/configure.ts
```diff
@@ -116,6 +116,69 @@ class Orchestrator {
...
+            case TargetResourceType.AKS:
+                let templateParameterHelper = new TemplateParameterHelper();
+                let template = this.inputs.potentialTemplates[0] as LocalPipelineTemplate;
+                inputs.pipelineConfiguration.template = this.inputs.potentialTemplates[0];
+                await templateParameterHelper.setParameters(template.parameters, inputs);
+                break;
+
+            case TargetResourceType.WebApp:
+                var shortlistedTemplates = [];
+                shortlistedTemplates = this.inputs.potentialTemplates.filter((template) => template.targetKind === resource.kind);
```
bishal-pdMSFT, line 148: template.targetKind can be null or empty as well

bishal-pdMSFT, line 148: I did not see code to merge windows and linux templates, then so we need to do this filtering here?

src/configure/configure.ts
```diff
@@ -116,6 +116,69 @@ class Orchestrator {
...
+                if (shortlistedTemplates.length > 1) {
+                    inputs.pipelineConfiguration.template = shortlistedTemplates.find((template) => template.templateType === TemplateType.remote);
```
bishal-pdMSFT, line 150: this can be undefined if no remote template is found. In such case may be use first element. Please handle such corner cases.

src/configure/configure.ts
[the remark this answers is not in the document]
bishal-pdMSFT: is this here for windows web app? By the way Nitin has started work to add templates in template service for windows web app. So, may be you can assume that both linux and windows web app flows will use remote templates. Will that help simplify the code?

src/configure/configure.ts
```diff
@@ -124,46 +187,21 @@ class Orchestrator {
...
+                await this.selectTemplate(selectedResource, this.inputs);
+            }
+            else {
+                await this.selectTemplate(resourceNode, this.inputs);
+            }
 
-
-            if (this.inputs.pipelineConfiguration.templateInfo) {
-                let extendedPipelineTemplate = await templateHelper.getTemplateParameteres(this.inputs.azureSession, this.inputs.pipelineConfiguration.templateInfo);
+            if (this.inputs.pipelineConfiguration.template.templateType === TemplateType.remote) {
```
bishal-pdMSFT, line 198: Move this to a separate function so that getInputs() function is pretty easy to understand. And you can move the logic for populating inputs for AKS flow to else part here

src/configure/configure.ts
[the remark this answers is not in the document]
kanika1894: The code is written in such a manner that if there are templates in both local and remote for same app, then remote one will be preferred.

```
  var shortlistedTemplates = [];
                shortlistedTemplates = this.inputs.potentialTemplates.filter((template) => template.targetKind === resource.kind);
                if (shortlistedTemplates.length > 1) {
                    this.inputs.pipelineConfiguration.template = shortlistedTemplates.find((template) => template.templateType === TemplateType.REMOTE);
                }
                else {
                    this.inputs.pipelineConfiguration.template = shortlistedTemplates[0];
                }
                break;
```

src/configure/configure.ts
kanika1894: yeah, now everything is in place, so have removed the dummy result

src/configure/configure.ts
kanika1894: Yes, analyzeRepoAndListAppropriatePipeline2() merges both local and remote templates.

src/configure/configurers/AksAzureResourceSelector.ts
kanika1894: For AKS, everything is done in a separate function called setParameters which we have reused as it is.  AKS resources dropdown is shown to the user while processing the predefined AKS template.

src/configure/configurers/ResourceSelectorFactory.ts
kanika1894: The label is decided. Potential templates will have all the labels having the same label.

src/configure/helper/templateHelper.ts
kanika1894: Code updated. Even if the call to remote templates fails due to any reason, it will still result the list of local templates.

src/configure/model/models.ts
kanika1894: The list of potential templates stores the templates (both local and remote as well as both linux/windows, depending upon the user selection of the template label). Later, when the user selects the resource, we decide the final template.

src/configure/helper/templateHelper.ts
kanika1894: umm.. let's decide the weights for the local templates, and I'll update all the templates at once?

src/configure/configure.ts
kanika1894: The potentialTemplates array is having merged templates = linux, windows and both local, remote. As this stage, we are deciding the template based on what resource did the user select(TargetKind) and what all were supported in the template service(TemplateType).

src/configure/configure.ts
kanika1894: yes, I guess this is a part of a separate PR by @anuragc617

src/configure/configure.ts
anuragc617: I think analyzeRepoAndListAppropriatePipeline2 will return the github remote templates + Azure local templates in case of azure repos.

kanika1894: Assuming that we'll be calling template service for github templates only, have added the check for repoProvider.

src/configure/helper/templateHelper.ts
kanika1894: updated

src/configure/configure.ts
kanika1894: For a web app scenario, for remote templates, we have defined targetKinds as follows :
```
export function getTargetKind(template: PipelineTemplateMetadata) {
    var targetKind: TargetKind;
    if (template.attributes.deployTarget.toLowerCase().includes("webapp")) {
        if (template.attributes.deployTarget.toLowerCase().includes("container")) {
            targetKind = TargetKind.LinuxContainerApp;
        } else if (template.attributes.deployTarget.toLowerCase().includes("linux")) {
            targetKind = TargetKind.LinuxApp;
        } else if (template.attributes.deployTarget.toLowerCase().includes("windows")) {
            targetKind = TargetKind.WindowsApp;
        }
    }
    else {
        targetKind = null;
    }
    return targetKind;
}
```

And for local templates, only "Build and Test Python Django App" has targetKind as null. Not sure if it will be a part of webApp scenario.

src/configure/configure.ts
kanika1894: Such case would occur if we have multiple local templates for the same targetKind and don't have any in the remote template. Please help me a bit in understanding this possibility.

src/configure/model/templateModels.ts
```diff
@@ -1,26 +1,44 @@
...
+    local
+}
 export interface PipelineTemplate {
-    path: string;
     label: string;
-    language: string;
+    templateWeight: number;
+    templateType: TemplateType;
```
bishal-pdMSFT, line 10: I missed to find where this type is being set. Can you please point me out?

src/configure/model/templateModels.ts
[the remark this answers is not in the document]
bishal-pdMSFT: @anuragc617 I have not yet seen your PR for finding working directory, but I want to tell one thing - you should use the working directory property, which is returned along with template id from template service, for the purpose of figuring our working directory rather than using template attributes.
The reason is that template service has already computed which applicationSetting can get deployed using what templates. And that information it is exposing by returning the working directory property. So we should use this instead of trying ourselves to co-relate between applicationSettings and templates.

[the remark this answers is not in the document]
bishal-pdMSFT: cc @tejasd1990

src/configure/helper/templateHelper.ts
bishal-pdMSFT: @anuragc617 right now there is no plan to add azure pipelines templates. But lets not take dependency on this as things might change

src/configure/helper/templateHelper.ts
bishal-pdMSFT: We will always use remote template over a local template with same name. This fact does not change even if specific template is selected only after webapp is selected

src/configure/configure.ts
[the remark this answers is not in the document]
kanika1894: added a check for ``` this.continueOrchestration ```

src/configure/configurers/IAzureResourceSelector.ts
kanika1894: done

src/configure/model/templateModels.ts
kanika1894: For local templates, it's defined in the template. For remote, we are setting it in templateHelper.ts.
```
export function getTargetType(template: PipelineTemplateMetadata) {
    if (template.attributes.deployTarget.toLowerCase().includes("webapp")) {
        return TargetResourceType.WebApp;
    } else if (template.attributes.deployTarget === "Azure:AKS") {
        return TargetResourceType.AKS;
    }
    return TargetResourceType.None;
}
```

src/configure/configure.ts
bishal-pdMSFT: The possibility is almost zero. But none the less it is good idea to handle such corner cases explicitly and log appropriate error messages. Otherwise the error message would be too cryptic to understand the root cause without debugging.

src/configure/configure.ts
bishal-pdMSFT: When I say merged for windows and linux, it means that we change the template label to hide windows and linux words from there.
This merge is different from merging of local and remote templates. May be we should call it something different to be clear. morphed templates??

src/configure/model/templateModels.ts
bishal-pdMSFT: My question was for templateType and not targetType

src/configure/configure.ts
bishal-pdMSFT: The label is a morphed label. it maps to more than one template. The template is decided only after web app is chosen. and that's where the telemetry should be set.

kanika1894: Updated

src/configure/helper/templateHelper.ts
kanika1894: Have updated the weights to 100

src/configure/configure.ts
```diff
@@ -116,54 +116,75 @@ class Orchestrator {
...
+                if (shortlistedTemplates.length > 1) {
+                    this.inputs.pipelineConfiguration.template = shortlistedTemplates.find((template) => template.templateType === TemplateType.REMOTE);
+                }
+                else {
+                    this.inputs.pipelineConfiguration.template = shortlistedTemplates[0];
+                }
+                break;
+
+            default:
+                throw new Error("Resource not supported");
```
bishal-pdMSFT, line 158: I think we define all error messages in a different file

src/configure/helper/templateHelper.ts
anuragc617: We can't filter local templates even if the name is same at this stage because local template can be for different kind of web App like windows and remote templates may not contain windows webapp tempates. We will get to know the webAppKind after the webApp selection.

src/configure/configurers/ResourceSelectorFactory.ts
```diff
@@ -0,0 +1,18 @@
...
+export class ResourceSelectorFactory {
+
+    public static getAzureResourceSelector(targetType: TargetResourceType): IAzureResourceSelector {
+        switch (targetType) {
+            case TargetResourceType.WebApp:
+                return new WebAppAzureResourceSelector();
+            case TargetResourceType.AKS:
+                return new AksAzureResourceSelector();
+            default:
+                throw new Error("Not supported");
```
anuragc617, line 15: define proper error message in Message.ts and use it here.

src/configure/helper/templateHelper.ts
```diff
@@ -18,9 +18,9 @@ export async function mergingRepoAnalysisResults(repoPath: string, repositoryPro
     let analysisResult = localRepoAnalysisResult;
 
     //If Repo analysis fails then we'll go with the basic existing analysis
-    if (repositoryProvider === RepositoryProvider.Github && !!repoAnalysisParameters && !!repoAnalysisParameters.repositoryAnalysisApplicationSettingsList) {
+    if (repositoryProvider === RepositoryProvider.Github && !!repoAnalysisParameters && !!repoAnalysisParameters.analysisApplicationSettingsList) {
```
anuragc617, line 21: it should be applicationSettingsList not analysisApplicationSettingsList.

src/configure/helper/templateHelper.ts
```diff
@@ -41,11 +41,42 @@ export async function mergingRepoAnalysisResults(repoPath: string, repositoryPro
     }
     return analysisResult;
 }
-export async function analyzeRepoAndListAppropriatePipeline(repoPath: string, repositoryProvider: RepositoryProvider, repoAnalysisParameters: RepositoryAnalysisParameters, targetResource?: GenericResource): Promise<PipelineTemplate[]> {
+
+export function getTargetType(template: PipelineTemplateMetadata) {
```
anuragc617, line 45: add return type.

src/configure/helper/templateHelper.ts
```diff
@@ -41,11 +41,42 @@ export async function mergingRepoAnalysisResults(repoPath: string, repositoryPro
...
+    if (template.attributes.deployTarget.toLowerCase().includes("webapp")) {
+        return TargetResourceType.WebApp;
+    } else if (template.attributes.deployTarget === "Azure:AKS") {
+        return TargetResourceType.AKS;
+    }
+    return TargetResourceType.None;
+}
+
+export function getTargetKind(template: PipelineTemplateMetadata) {
```
anuragc617, line 54: add return type of function here as well.

src/configure/model/models.ts
```diff
@@ -35,10 +35,11 @@ export class WizardInputs {
...
 }
 
 export class RepositoryAnalysisParameters {
-    repositoryAnalysisApplicationSettingsList: RepositoryAnalysisApplicationSettings[];
+    analysisApplicationSettingsList: RepositoryAnalysisApplicationSettings[];
```
anuragc617, line 42: it should be applicationSettingsList.

src/configure/model/templateModels.ts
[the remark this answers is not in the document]
anuragc617: @bishal-pdMSFT template's working directory can match to multiple repositoryAnalysisSettings workingDirectory. So to narrow down more on repositoryAnalysisSetting, filtering should be done on templates deployTarget, buildTarget  and language attributes.

src/configure/configure.ts
kanika1894: Since discussed on call, resolving it

src/configure/configure.ts
kanika1894: Added the required check

src/configure/configure.ts
kanika1894: done

src/configure/configurers/ResourceSelectorFactory.ts
kanika1894: updated

src/configure/helper/templateHelper.ts
kanika1894: updated

src/configure/helper/templateHelper.ts
kanika1894: added

src/configure/helper/templateHelper.ts
kanika1894: added

src/configure/model/models.ts
kanika1894: updated

src/configure/configure.ts
```diff
@@ -116,54 +116,75 @@ class Orchestrator {
...
         if (this.continueOrchestration) {
             await this.getSourceRepositoryDetails();
             await this.getAzureSession();
             await this.getGithubPatToken();
-            await this.getSelectedPipeline();
-
+            var targetType = await this.getSelectedPipeline();
+            if (!resourceNode) {
+                let selectedResource = await this.getAzureResource(targetType);
+                if (!!selectedResource && this.continueOrchestration) {
```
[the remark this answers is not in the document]
kanika1894: updated

PR #85, opened 2020-04-27
Repository analysis setting detection and integration with template parameters.

anuragc617:


src/configure/configure.ts
```diff
@@ -516,36 +514,49 @@ class Orchestrator {
...
+        //If RepoAnalysis is disabled or didn't provided response related to language of selected template
+        this.inputs.repositoryAnalysisApplicationSettings = new RepositoryAnalysisApplicationSettings();
+
+        if (!repoAnalysisResult || !repoAnalysisResult.applicationSettingsList) {
+            return;
+        }
+        var applicationSettings = repoAnalysisResult.applicationSettingsList.filter(applicationSetting => {
+            if (this.inputs.pipelineConfiguration.template.templateType === TemplateType.remote) {
+                const template = this.inputs.pipelineConfiguration.template as RemotePipelineTemplate;
+                if (applicationSetting.language && applicationSetting.language !== template.attributes.language.toLowerCase()) {
```
bishal-pdMSFT, line 530: Rather than finding out applicationSettings by comparing these properties, a better way is to use workingDirectory which is returned by template service (may be store WD in RemotePIpelineTemplate object if not already doing so. cc @tejasd1990 for helping here

anuragc617: I will add workingDirectory check also. But to do more filtering of repositoryAnalysisSettings we need to compare these properties as well. Consider a scenarios where a workingDirectory contains two types of application settings ex: gruntFile settings and DockerFile settings and suppose user has selected DockerFile template. In this case as the workingDirectory will be same for gruntFile settings and DockerFile settings, we won't be able to filter on workingDirectory alone, we will have to filter by comparing application setting (language, deployTarget and buildTarget).

src/configure/configure.ts
```diff
@@ -118,37 +119,21 @@ class Orchestrator {
...
                 break;
-
             case TargetResourceType.WebApp:
                 var shortlistedTemplates = [];
                 shortlistedTemplates = this.inputs.potentialTemplates.filter((template) => template.targetKind === resource.kind);
                 if (!!shortlistedTemplates && shortlistedTemplates.length > 1) {
                     this.inputs.pipelineConfiguration.template = shortlistedTemplates.find((template) => template.templateType === TemplateType.REMOTE);
-                } else if (!!shortlistedTemplates) {
+                }
+                else if (!!shortlistedTemplates) {
```
bishal-pdMSFT, line 136: why move else to new line? If I know correctly, else should be in same line as preceding closing brace for nodejs/javascript. No?

src/configure/configure.ts
```diff
@@ -166,29 +151,43 @@ class Orchestrator {
         if (this.continueOrchestration) {
             await this.getSourceRepositoryDetails();
             await this.getAzureSession();
-            await this.getGithubPatToken();
-            var targetType = await this.getSelectedPipeline();
-            if (!resourceNode) {
-                let selectedResource = await this.getAzureResource(targetType);
-                if (!this.continueOrchestration) {
+            let repoAnalysisResult = await this.getRepositoryAnalysis();
+            let targetType = await this.getSelectedPipeline(repoAnalysisResult);
```
bishal-pdMSFT, line 155: looks wierd that `getSelectedPipeline` returns targetType. Rather it should return void and there should be another function like `getSelectedPipelineTargetType()` which just does `return this.inputs.potentialTemplates[0].targetType`

src/configure/configure.ts
```diff
@@ -509,21 +500,35 @@ class Orchestrator {
...
 
         if (!applicationSettings || applicationSettings.length === 0) {
             return;
         }
-
-        let workspacePaths = Array.from(new Set(applicationSettings.map(a => a.settings.workingDirectory)));
-        if (workspacePaths.length === 1) {
+        this._repoAnalysisSettings = applicationSettings;
+        if (applicationSettings.length === 1) {
             this.inputs.repositoryAnalysisApplicationSettings = applicationSettings[0];
```
bishal-pdMSFT, line 522: I assume this is only used by local templates

src/configure/helper/repoAnalysisHelper.ts
```diff
@@ -79,15 +79,18 @@ export class RepoAnalysisHelper {
                                     applicationSettings.settings.workingDirectory, analysis.settings[RepoAnalysisConstants.RequirementsFilePath]);
                                 applicationSettings.settings.pythonRequirementsFileDirectory = path.dirname(analysis.settings[RepoAnalysisConstants.RequirementsFilePath]);
                             }
+                        } else {
+                            applicationSettings.settings = analysis.settings;
```
bishal-pdMSFT, line 83: when is this condition useful?

src/configure/helper/templateHelper.ts
```diff
@@ -162,11 +162,9 @@ export async function analyzeRepoAndListAppropriatePipeline2(azureSession: Azure
                     label: template.templateDescription,
                     targetType: getTargetType(template),
                     targetKind: getTargetKind(template),
-                    language: template.attributes.language,
-                    templateId: template.templateId,
-                    templateWeight: template.templateWeight,
                     templateType: TemplateType.REMOTE,
-                    workingDirectory: template.workingDirectory
+                    language: template.attributes.language,
+                    ...template
```
bishal-pdMSFT, line 167: what is the reason for needing this?

src/configure/templateInputHelper/InputControl.ts
```diff
@@ -72,10 +72,10 @@ export class InputControl {
                 this.value = await this.dataSource.evaluateDataSources(dependentInputs, azureSession);
             }
             else if (this.controlType === ControlType.QuickPick) {
-                let selectedValue = await this.dataSource.evaluateDataSources(dependentInputs, azureSession)
-                    .then((listItems: Array<{ label: string, data: any, group?: string }>) => {
-                        return this.showQuickPick(this.getInputControlId(), listItems, { placeHolder: this.inputDescriptor.name });
-                    });
+                let selectedValue = await this.showQuickPick(this.getInputControlId(),
```
bishal-pdMSFT, line 75: If I am not wrong, this was a window notification. Are you changing to quickpick?

src/configure/templateInputHelper/InputControlProvider.ts
```diff
@@ -1,26 +1,36 @@
...
 
 export class InputControlProvider {
+    private readonly repoAnalysisSettingKey: string = "repoAnalysisSettingKey";
     private _pipelineTemplate: ExtendedPipelineTemplate;
     private _inputControlsMap: Map<string, InputControl>;
+    private _repoAnalysisSettings: RepositoryAnalysisApplicationSettings[];
+    private _repoAnalysisSettingInUse: number;
 
-    constructor(pipelineTemplate: ExtendedPipelineTemplate, context: { [key: string]: any }) {
+    constructor(pipelineTemplate: ExtendedPipelineTemplate, repoAnalysisSettings: RepositoryAnalysisApplicationSettings[], context: { [key: string]: any }) {
```
bishal-pdMSFT, line 18: can repoAnalysisSettings be part of context?

src/configure/templateInputHelper/InputControlProvider.ts
```diff
@@ -63,6 +73,50 @@ export class InputControlProvider {
         this._setInputControlDataSourceInputs();
     }
 
+    private async setInputControlValueFromRepoAnalysisResult(inputControl: InputControl): Promise<void> {
+        let repoAnalysisSettingKey = inputControl.getPropertyValue(this.repoAnalysisSettingKey);
+        if (this._repoAnalysisSettingInUse !== -1) {
```
bishal-pdMSFT, line 78: the usage of _repoAnalysisSettingInUse  is really tricky and overloaded. it is both used as a boolean in if condition and also as an index in array. this is not so good.

bishal-pdMSFT: can you articulate in simple terms what is intention behind using _repoAnalysisSettingInUse

src/configure/templateInputHelper/InputControlProvider.ts
```diff
@@ -63,6 +73,50 @@ export class InputControlProvider {
...
+                    return;
+                }
+                if (settingIndexMap.has(analysisSetting.settings[repoAnalysisSettingKey])) {
+                    settingIndexMap.get(analysisSetting.settings[repoAnalysisSettingKey]).push(index);
+                } else {
+                    settingIndexMap.set(analysisSetting.settings[repoAnalysisSettingKey], [index]);
+                }
+            });
+            if (settingIndexMap.size === 0) {
+                let value = await new ControlProvider().showInputBox(repoAnalysisSettingKey, { placeHolder: inputControl.getInputDescriptor().name });
```
bishal-pdMSFT, line 98: please log a trace here as ideally this is not expected

src/configure/templateInputHelper/InputControlProvider.ts
```diff
@@ -63,6 +73,50 @@ export class InputControlProvider {
...
+        let repoAnalysisSettingKey = inputControl.getPropertyValue(this.repoAnalysisSettingKey);
+        if (this._repoAnalysisSettingInUse !== -1) {
+            if (this._repoAnalysisSettings.length === 0 || !this._repoAnalysisSettings[this._repoAnalysisSettingInUse].settings[repoAnalysisSettingKey]) {
+                let value = await new ControlProvider().showInputBox(repoAnalysisSettingKey, { placeHolder: inputControl.getInputDescriptor().name });
+                inputControl.setValue(value);
+            } else {
+                inputControl.setValue(this._repoAnalysisSettings[this._repoAnalysisSettingInUse].settings[repoAnalysisSettingKey]);
+            }
+        } else {
+            let settingIndexMap: Map<string, number[]> = new Map();
```
bishal-pdMSFT, line 86: what is the need for a map with an number array value?

src/configure/templateInputHelper/InputControlProvider.ts
```diff
@@ -63,6 +73,50 @@ export class InputControlProvider {
...
+
+                if (possibleValues.length === 1) {
+                    selectedValue = possibleValues[0];
+                }
+                else {
+                    selectedValue = await new ControlProvider().showQuickPick(repoAnalysisSettingKey, possibleValues, { placeHolder: inputControl.getInputDescriptor().name });
+                }
+
+                if (settingIndexMap.get(selectedValue.data).length === 1) {
+                    this._repoAnalysisSettingInUse = settingIndexMap.get(selectedValue.data)[0];
```
bishal-pdMSFT, line 113: okk now I understand why you are using _repoAnalysisSettingInUse

bishal-pdMSFT: the name is confusing. should we call it _selectedRepoAnalysisSettingIndex ?

src/configure/templateInputHelper/InputControlProvider.ts
```diff
@@ -1,26 +1,36 @@
...
     private _pipelineTemplate: ExtendedPipelineTemplate;
     private _inputControlsMap: Map<string, InputControl>;
+    private _repoAnalysisSettings: RepositoryAnalysisApplicationSettings[];
+    private _repoAnalysisSettingInUse: number;
 
-    constructor(pipelineTemplate: ExtendedPipelineTemplate, context: { [key: string]: any }) {
+    constructor(pipelineTemplate: ExtendedPipelineTemplate, repoAnalysisSettings: RepositoryAnalysisApplicationSettings[], context: { [key: string]: any }) {
         this._pipelineTemplate = pipelineTemplate;
         this._inputControlsMap = new Map<string, InputControl>();
+        this._repoAnalysisSettings = repoAnalysisSettings || [];
```
bishal-pdMSFT, line 21: Can all repo analysis related fields and methods be moved to a separate class? this will clean IPC class cleaner.

src/configure/templateInputHelper/InputControlProvider.ts
```diff
@@ -63,6 +73,50 @@ export class InputControlProvider {
         this._setInputControlDataSourceInputs();
     }
 
+    private async setInputControlValueFromRepoAnalysisResult(inputControl: InputControl): Promise<void> {
+        let repoAnalysisSettingKey = inputControl.getPropertyValue(this.repoAnalysisSettingKey);
+        if (this._repoAnalysisSettingInUse !== -1) {
+            if (this._repoAnalysisSettings.length === 0 || !this._repoAnalysisSettings[this._repoAnalysisSettingInUse].settings[repoAnalysisSettingKey]) {
```
bishal-pdMSFT, line 79: should we not be even inside setInputControlValueFromRepoAnalysisResult function if `this._repoAnalysisSettings.length === 0`. this will keep the code cleaner

src/configure/templateInputHelper/InputControlProvider.ts
```diff
@@ -63,6 +73,50 @@ export class InputControlProvider {
         this._setInputControlDataSourceInputs();
     }
 
+    private async setInputControlValueFromRepoAnalysisResult(inputControl: InputControl): Promise<void> {
```
bishal-pdMSFT, line 76: As suggested earlier it would be good to de-couple getting settings values from repo analysis and showing input control. A separate class can take care of returning matched analysis setting(s) and then ICP class can draw input based on count of retuned settings (and handle null as well).

bishal-pdMSFT:
This is really good work @anuragc617

src/configure/templateInputHelper/InputControlProvider.ts
anuragc617: Done

src/configure/templateInputHelper/InputControlProvider.ts
anuragc617: done

src/configure/templateInputHelper/InputControlProvider.ts
anuragc617: created a new class RepoAnalysisSettingInputProvider for this.

src/configure/templateInputHelper/InputControlProvider.ts
anuragc617: done

src/configure/templateInputHelper/InputControlProvider.ts
anuragc617: done

src/configure/configure.ts
anuragc617: done

PR #86, opened 2020-04-28
Consuming contracts package for Repository Analysis

issacnitin:


package-lock.json
```diff
@@ -799,6 +799,21 @@
...
         },
+        "azureintegration_repoanalysis_client_internal": {
+            "version": "1.0.0",
+            "resolved": "https://registry.npmjs.org/azureintegration_repoanalysis_client_internal/-/azureintegration_repoanalysis_client_internal-1.0.0.tgz",
+            "integrity": "sha512-GCkedG6EHbN7c9QfmiBlg3VW64/F4J2BNcMIkurwTgXtejQrgv611iYR0jqFEwyMy9Nj14fkRxFB94ApjdH23g==",
+            "requires": {
+                "typescript": "^3.8.3"
+            },
+            "dependencies": {
+                "typescript": {
```
bishal-pdMSFT, line 810: typescript should be a dev-dependency. can you please fix this in the package?

package-lock.json
```diff
@@ -799,6 +799,21 @@
                 "ms-rest-azure": "^2.5.5"
             }
         },
+        "azureintegration_repoanalysis_client_internal": {
```
bishal-pdMSFT, line 802: do we have `_` in package name? I think it should be `-`. I am not really sure if this could be a problem as I have not read so anywhere but lets just follow the common practice.

PR #87, opened 2020-05-04
Added static and dynamic validations for tempate parameters

anuragc617:


src/configure/clients/azure/armRestClient.ts
```diff
@@ -10,7 +10,7 @@ export class ArmRestClient {
         this.restClient = new RestClient(azureSession.credentials);
     }
 
-    public fetchArmData(endPointUri: string, httpMethod: string, body?: string) {
+    public fetchArmData(endPointUri: string, httpMethod: string, body?: any) {
```
bishal-pdMSFT, line 13: why this change?

src/configure/configure.ts
```diff
@@ -446,7 +446,7 @@ class Orchestrator {
             repoAnalysisResult = await repoAnalysisHelper.getRepositoryAnalysis(this.inputs.sourceRepository, this.inputs.pipelineConfiguration.workingDirectory.split('\\').join('/'));
         }
 
-        extensionVariables.templateServiceEnabled = false;
+        extensionVariables.templateServiceEnabled = true;
```
bishal-pdMSFT, line 449: please turn it off

src/configure/configure.ts
```diff
@@ -513,7 +513,7 @@ class Orchestrator {
         }
 
         if (extensionVariables.templateServiceEnabled) {
-            telemetryHelper.setTelemetry(TelemetryKeys.ChosenTemplate, this.inputs.pipelineConfiguration.templateInfo.label);
+            telemetryHelper.setTelemetry(TelemetryKeys.ChosenTemplate, this.inputs.pipelineConfiguration.templateInfo.templateDescription);
```
bishal-pdMSFT, line 516: this might cause an issue if we want to look at both current and historical data. the format is changing.
Should we rather log `<label>:<targetKind>` ?

src/configure/model/templateModels.ts
```diff
@@ -17,10 +17,10 @@ export interface PipelineTemplate {
 
 export interface PipelineTemplateMetadata {
     templateId: string;
-    label: string;
+    templateDescription: string;
```
bishal-pdMSFT, line 20: Okay I see you have renamed this. But while logging telemetry, I still think you should append target kind if it exists

bishal-pdMSFT, line 20: just `description`

src/configure/templateInputHelper/InputControl.ts
```diff
@@ -62,21 +73,31 @@ export class InputControl {
...
             var dependentInputs = this._getDataSourceInputs();
             if (this.controlType === ControlType.None || this.controlType === ControlType.InputBox) {
-                this.value = await this.dataSource.evaluateDataSources(dependentInputs, azureSession);
+                this.value = await this.dataSource.evaluateDataSources(dependentInputs, this.azureSession);
             }
             else if (this.controlType === ControlType.QuickPick) {
-                let selectedValue = await this.dataSource.evaluateDataSources(dependentInputs, azureSession)
-                    .then((listItems: Array<{ label: string, data: any, group?: string }>) => {
-                        return this.showQuickPick(this.getInputControlId(), listItems, { placeHolder: this.inputDescriptor.name });
+                let listItems: Array<QuickPickItemWithData> = await vscode.window.withProgress(
```
bishal-pdMSFT, line 86: Nice

bishal-pdMSFT, line 86: Nit: Should the return type be QuickPickItemWithData or something more generic?

src/configure/templateInputHelper/InputControl.ts
```diff
@@ -62,21 +73,31 @@ export class InputControl {
...
+                    { location: vscode.ProgressLocation.Notification, title: utils.format(Messages.fetchingInputMessage, this.inputDescriptor.name) },
```
bishal-pdMSFT, line 87: question: what other location is possible?

src/configure/templateInputHelper/InputControl.ts
```diff
@@ -62,21 +73,31 @@ export class InputControl {
...
-    public async setInputControlValue(azureSession: AzureSession): Promise<any> {
+    public async setInputControlValue(): Promise<any> {
         if (!this.isVisible()) {
             return;
         }
         if (!!this.dataSource) {
             var dependentInputs = this._getDataSourceInputs();
             if (this.controlType === ControlType.None || this.controlType === ControlType.InputBox) {
-                this.value = await this.dataSource.evaluateDataSources(dependentInputs, azureSession);
+                this.value = await this.dataSource.evaluateDataSources(dependentInputs, this.azureSession);
```
bishal-pdMSFT, line 83: Won't we need to trigger dynamic/static validation for input box?

bishal-pdMSFT: ahh okay we are only getting the value here

src/configure/templateInputHelper/InputControl.ts
```diff
@@ -95,10 +116,9 @@ export class InputControl {
             }
             else if (this.controlType === ControlType.InputBox) {
                 this.value = await this.showInputBox(this.getInputControlId(), {
+                    value: this.value,
```
bishal-pdMSFT, line 119: why is this change required?

src/configure/templateInputHelper/InputControlProvider.ts
```diff
@@ -1,58 +1,62 @@
...
 import { VisibilityHelper } from "./utilities/VisibilityHelper";
 
 export class InputControlProvider {
     private _pipelineTemplate: ExtendedPipelineTemplate;
     private _inputControlsMap: Map<string, InputControl>;
+    private azureSession: AzureSession;
+    private _context: StringMap<any>;
 
-    constructor(pipelineTemplate: ExtendedPipelineTemplate, context: { [key: string]: any }) {
+    constructor(azureSession: AzureSession, pipelineTemplate: ExtendedPipelineTemplate, context: { [key: string]: any }) {
```
bishal-pdMSFT, line 16: use StringMap here as well?

src/configure/templateInputHelper/InputControlProvider.ts
```diff
@@ -1,58 +1,62 @@
...
+        this._createControls();
     }
 
-    public async getAllPipelineTemplateInputs(azureSession: AzureSession, resourceNode?: GenericResource) {
+    public async getAllPipelineTemplateInputs() {
         let parameters: { [key: string]: any } = {};
         for (let inputControl of this._inputControlsMap.values()) {
-            if (inputControl.getPropertyValue('deployTarget') === "true" && !!resourceNode) {
-                inputControl.setValue(resourceNode.id);
+            if (inputControl.getPropertyValue('deployTarget') === "true" && !!this._context["resourceId"]) {
```
bishal-pdMSFT, line 27: using an index into a map can lead to very cryptic error messages like `element not found`. you can define a function like `getItem(key: string)` which will return a value or will throw a friendly error message

bishal-pdMSFT, line 27: If I am not wrong, dependency on this property `deployTarget` is subject to change based on how we want to surface this information in template. right?

src/configure/templateInputHelper/InputControlProvider.ts
```diff
@@ -81,6 +86,42 @@ export class InputControlProvider {
...
+                    var validationDataSource = DataSourceUtility.getDataSourceById(this._pipelineTemplate.dataSources, validation.dataSourceId);
+                    if (validationDataSource) {
+                        dataSourceToInputsMap.set(validationDataSource, []);
+
+                        let dynamicValidationRequiredInputIds = DataSourceUtility.getDependentInputIdList(validationDataSource.endpointUrlStem);
+                        dynamicValidationRequiredInputIds = dynamicValidationRequiredInputIds.concat(DataSourceUtility.getDependentInputIdList(validationDataSource.requestBody));
+                        dynamicValidationRequiredInputIds = dynamicValidationRequiredInputIds.concat(DataSourceUtility.getDependentInputIdList(validationDataSource.resultTemplate));
+                        dynamicValidationRequiredInputIds = dynamicValidationRequiredInputIds.concat(DataSourceUtility.getDependentInputIdList(validationDataSource.resultSelector));
+
+                        dynamicValidationRequiredInputIds = Array.from(new Set(dynamicValidationRequiredInputIds));
```
bishal-pdMSFT, line 106: why do we need this Array.from ?

src/configure/templateInputHelper/InputControlProvider.ts
```diff
@@ -81,6 +86,42 @@ export class InputControlProvider {
...
+                        dynamicValidationRequiredInputIds.forEach((dynamicValidationRequiredInputId) => {
+                            var dependentInput = this._inputControlsMap.get(dynamicValidationRequiredInputId);
+                            if (dependentInput && dynamicValidationRequiredInputId !== inputDes.id) {
+                                dataSourceToInputsMap.get(validationDataSource).push(dependentInput);
+                            } else {
+                                //.logError("Dependent input {0} specified for input {1} is not present in pipeline template {2}".format(dynamicValidationRequiredInputId, inputDes.id, this._pipelineTemplate.id));
```
bishal-pdMSFT, line 112: we can log trace for this

src/configure/templateInputHelper/InputControlProvider.ts
```diff
@@ -81,6 +86,42 @@ export class InputControlProvider {
...
+                            }
+                        });
+                    } else {
+                        //PipelineTemplateErrorLogger.logError("validation data source {0} specified for input {1} is not present in pipeline template {2}".format(validation.dataSourceId, inputDes.id, this._pipelineTemplate.id));
```
bishal-pdMSFT, line 116: we can log trace for this

bishal-pdMSFT: actually we should log only if `dependentInput` is null.

bishal-pdMSFT, line 116: it would be good to trace these errors, as they help in figuring out if there are some issues in our templates

src/configure/helper/mustacheHelper.ts
```diff
@@ -95,6 +95,25 @@ export class MustacheHelper {
                     }
 
                     return "";
+                };
+            },
+            "equals": function () {
+                return function (text: string, render: any) {
```
kanika1894, line 101: As we discussed, I have already implemented equals in my rightClickScenario PR

src/configure/configure.ts
kanika1894: This seems to be pretty old code as ``` pipelineConfiguration.templateInfo ``` doesn't exist anymore.

src/configure/templateInputHelper/InputControlProvider.ts
kanika1894: Yes, this part is changed as now we'll be using the concept of override in properties. This was initially done to handle right click scenario.

src/configure/clients/github/TemplateServiceClient.ts
```diff
@@ -20,7 +19,7 @@ export class TemplateServiceClient {
             body);
     }
 
-    public async getTemplateParameters(templateId: string): Promise<ExtendedPipelineTemplate> {
+    public async getTemplateParameters(templateId: string): Promise<any> {
```
kanika1894, line 22: Ideally we should have a fix return type, right? so, any specific reason for this change?

src/configure/templateInputHelper/InputControl.ts
```diff
@@ -123,6 +173,92 @@ export class InputControl {
...
+                    validationResult = validationResult.concat(result);
+                }
+            }
+        }
+        return validationResult;
+    }
+
+    private async evaluateDynamicValidation(validation: InputDynamicValidation, value: string): Promise<string> {
+        var requiredDataSource: DataSource = null;
+        this.validationDataSourceToInputsMap.forEach((descriptorsArray, dataSource) => { if (dataSource.id === validation.dataSourceId) { requiredDataSource = dataSource; } });
```
bishal-pdMSFT, line 211: add types for descriptorsArray and dataSource

bishal-pdMSFT, line 211: please format the code

bishal-pdMSFT: in multiple lines

bishal-pdMSFT, line 211: Shouldn't the order of arguments be reversed - (dataSource, descriptorsArray) as datasource is key here?

src/configure/templateInputHelper/InputControl.ts
```diff
@@ -110,11 +130,41 @@ export class InputControl {
...
         if (!!properties) {
             value = properties[propertyName];
-            if( !!value && !!inputs ) { return MustacheHelper.render(value, { inputs: inputs }); }
+            if (!!value && !!inputs) { return MustacheHelper.render(value, { inputs: inputs }); }
         }
         return value;
     }
 
+    public setValidationDataSources(dataSourceToInputsMap: Map<DataSource, InputControl[]>) {
+        this.validationDataSourceToInputsMap = dataSourceToInputsMap;
```
bishal-pdMSFT, line 139: Do we really need this map? What purpose does it serve? Should we just a list of <DataSource> as dynamic validations for an input are already stored in `this.inputDescriptor.dynamicValidations`

src/configure/templateInputHelper/InputControl.ts
```diff
@@ -110,11 +130,41 @@ export class InputControl {
...
+    }
+
+    public async triggerControlValueValidations(value: string): Promise<string> {
+        this._tempValue = value;
```
bishal-pdMSFT, line 143: temp is an ambiguous word. If I understand correctly it is the value which has been selected by user, has not yet been validated. If so, should we call it _valuePendingValidation  or something similar?

src/configure/templateInputHelper/utilities/DataSourceUtility.ts
```diff
@@ -39,7 +40,7 @@ export class DataSourceUtility {
         var httpMethod = dataSource.httpMethod || "GET";
         var requestBody = !!dataSource.requestBody ? MustacheHelper.render(dataSource.requestBody, view) : null;
         let amrClient = new ArmRestClient(azureSession);
-        return amrClient.fetchArmData(armUri, httpMethod, requestBody)
+        return amrClient.fetchArmData(armUri, httpMethod, JSON.parse(requestBody))
```
bishal-pdMSFT, line 43: why parse?

src/configure/clients/github/TemplateServiceClient.ts
bishal-pdMSFT: +1

anuragc617: While testing with default inputs I changed it, forgot to change back.

src/configure/configure.ts
anuragc617: done

src/configure/configure.ts
anuragc617: yes it doesn't exist, this branch was not updated after your changes.

src/configure/templateInputHelper/InputControl.ts
anuragc617: yes, changed it.

src/configure/templateInputHelper/InputControl.ts
anuragc617: Done

anuragc617: done

anuragc617: foreach in typescript has reverse order for Map -> (value, key)

src/configure/templateInputHelper/InputControlProvider.ts
anuragc617: because dynamicValidationRequiredInputIds is of Array type.

src/configure/templateInputHelper/utilities/DataSourceUtility.ts
anuragc617: MustacheHelper.render returns string, and the RestClient that we use for making API calls does stringify of body. If JSON object is not passed then it will become invalid string enclosed in two double quotes after stringify.

src/configure/templateInputHelper/InputControl.ts
anuragc617: this.inputDescriptor.dynamicValidations contains only dynamicValidations source Id, this is map of datasource  and required inputControls to evaluate the dynamic validation dataSource  which is set by InputControlProvider.

src/configure/templateInputHelper/InputControl.ts
anuragc617: thanks

src/configure/helper/mustacheHelper.ts
anuragc617: Needed to evaluate some dynamic validation conditions, removed it from my PR.

src/configure/templateInputHelper/InputControlProvider.ts
anuragc617: in typescript it return undefined.

src/configure/templateInputHelper/InputControl.ts
bishal-pdMSFT: What if the value returned by datasource is invalid? 
We should still run validation and if validation fails then show the input to user. What do you think? Otherwise user will get an error and they will not even know hot fix it.

src/configure/templateInputHelper/InputControlProvider.ts
bishal-pdMSFT: But `dynamicValidationRequiredInputIds` is also an array. No?

src/configure/templateInputHelper/InputControlProvider.ts
bishal-pdMSFT: I don't see any trace statement here

src/configure/templateInputHelper/InputControl.ts
bishal-pdMSFT: Got it.

src/configure/templateInputHelper/InputControlProvider.ts
anuragc617: here first a 'Set' is created to get only 'unique values' and then it is converted back to Array.

src/configure/templateInputHelper/InputControlProvider.ts
anuragc617: you looked at older commit, trace statement is there.

src/configure/templateInputHelper/InputControl.ts
anuragc617: yes that's right, but user may not know the correct value for it. But he should have the option to fix the value. Added it.

PR #93, opened 2020-05-19
Adding telemetry for new flow

kanika1894:


src/configure/configure.ts
```diff
@@ -453,44 +455,46 @@ class Orchestrator {
...
-                pipelineLabels.map((pipeline) => { return { label: pipeline }; }),
-                { placeHolder: Messages.selectPipelineTemplate },
-                TelemetryKeys.PipelineTempateListCount);
-            //only label gets finalized, template isn't final yet
-            this.inputs.potentialTemplates = pipelineMap.get(selectedOption.label);
-        }
-        else {
-            this.inputs.potentialTemplates = pipelineMap.get(pipelineLabels[0]);
-        }
+        await telemetryHelper.executeFunctionWithTimeTelemetry(async () => {
```
bishal-pdMSFT, line 458: This is too big scope. We can put the timer only around making call to template service. Rest are local processing and will be very quick.

src/configure/configure.ts
```diff
@@ -453,44 +455,46 @@ class Orchestrator {
...
+                        this.inputs.pipelineConfiguration.params[constants.TargetResource])
+                );
+                appropriatePipelines = remotePipelines;
+            }
+            let pipelineMap = this.getMapOfUniqueLabels(appropriatePipelines);
+            let pipelineLabels = Array.from(pipelineMap.keys());
+
+            // TO:DO- Get applicable pipelines for the repo type and azure target type if target already selected
+            if (pipelineLabels.length > 1) {
+                var selectedOption = await this.controlProvider.showQuickPick(
```
bishal-pdMSFT, line 486: Do not include this as it will include user's time to select a template.

src/configure/configure.ts
kanika1894: updated

kanika1894: okay, moving this to scope just the template service call

PR #94, opened 2020-05-20
Create asset and provisioning github workflow changes

anuragc617:


src/configure/clients/github/TemplateServiceClient.ts
```diff
@@ -28,4 +28,23 @@ export class TemplateServiceClient {
...
+        return this.restClient.sendRequest2(
+            requestUri,
+            'POST',
+            '6.0-preview.1',
+            inputs
+        );
+    }
+
+    public async getTemplateFile(templateId: string, fileId: string): Promise<string> {
+        let requestUri = this.templateServiceUri + "/" + templateId + "/files" + fileId;
```
bishal-pdMSFT, line 43: Does fileId contain a '/' at start? Should we put one here explicitly as well to ensure that url does not break

src/configure/configurers/configurerFactory.ts
```diff
@@ -1,14 +1,20 @@
...
 
 export class ConfigurerFactory {
-    public static GetConfigurer(sourceRepositoryDetails: GitRepositoryParameters, azureSession: AzureSession, subscriptionId: string): Configurer {
-        switch(sourceRepositoryDetails.repositoryProvider) {
+    public static GetConfigurer(sourceRepositoryDetails: GitRepositoryParameters, azureSession: AzureSession, subscriptionId: string, templateType: TemplateType, localGitRepoHelper: LocalGitRepoHelper): Configurer {
+        switch (sourceRepositoryDetails.repositoryProvider) {
             case RepositoryProvider.Github:
-                return extensionVariables.enableGitHubWorkflow ? new GitHubWorkflowConfigurer(azureSession, subscriptionId) : new AzurePipelineConfigurer(azureSession);
+                if (extensionVariables.enableGitHubWorkflow) {
+                    return templateType == TemplateType.LOCAL ? new GitHubWorkflowConfigurer(azureSession, subscriptionId) : new RemoteGitHubWorkflowConfigurer(azureSession, subscriptionId, localGitRepoHelper);
```
bishal-pdMSFT, line 15: should we rename `GitHubWorkflowConfigurer` to `LocalGitHubWorkflowConfigurer`

src/configure/model/models.ts
```diff
@@ -76,7 +76,6 @@ export class PipelineConfiguration {
     workingDirectory: string;
     params: { [key: string]: any } = {};
     assets: { [key: string]: any } = {};
-    parameters: { [key: string]: any } = {};
```
bishal-pdMSFT, line 79: what was the purpose of this property ?

src/configure/model/templateModels.ts
```diff
@@ -13,18 +15,17 @@ export interface PipelineTemplate {
...
     templateId: string;
     workingDirectory: string;
     templateWeight: number;
     templateDescription: string;
     attributes: TemplateAttributes;
 }
 
-export interface RemotePipelineTemplate extends PipelineTemplate {
-    templateId: string;
+export interface RemotePipelineTemplate extends PipelineTemplate, ExtendedPipelineTemplate {
```
bishal-pdMSFT, line 26: Just thinking loud about this. I am very clearly able to segregate what does this model represent. It is having three things:
1. ExtendedpipelineTemplate properties
2. Local pipeline template fills - templateType, targetKind etc
3. Working directory which is already part of templateInfo

I understand that we want to handle both local and remote template as same object and hence this class extends PipelineTemplate. So, basically this class represents all properties of extendedPipelineTemplate as well as some properties from local template.
But what I don't completely understand is why workingDirectory should be part of this. Working directory is already part of TemplateInfo. can we use that object?

src/configure/configure.ts
```diff
@@ -450,7 +462,7 @@ class Orchestrator {
     }
 
     private async getSelectedPipeline(repoAnalysisResult: RepositoryAnalysisParameters): Promise<void> {
-        extensionVariables.templateServiceEnabled = false;
+        extensionVariables.templateServiceEnabled = true;
```
bishal-pdMSFT, line 465: please turn if false while merging PR

src/configure/configure.ts
```diff
@@ -513,9 +525,12 @@ class Orchestrator {
         if (!repoAnalysisResult || !repoAnalysisResult.applicationSettingsList) {
             return;
         }
+        let workingDirectories = Array.from(new Set(this.inputs.potentialTemplates
+            .filter((template) => template.templateType === TemplateType.REMOTE)
+            .map((template: RemotePipelineTemplate) => template.workingDirectory)));
         var applicationSettings = repoAnalysisResult.applicationSettingsList.filter(applicationSetting => {
             if (this.inputs.pipelineConfiguration.template.templateType === TemplateType.REMOTE) {
-                return true;
+                return workingDirectories.indexOf(applicationSetting.settings.workingDirectory) > 0;
```
bishal-pdMSFT, line 533: `>= 0`

bishal-pdMSFT, line 533: Would it be better to handle casing here. Although not required currently, it is better to not take dependency on such facts.

src/configure/utilities/templateConverter.ts
```diff
@@ -0,0 +1,106 @@
+const templateParametersRegex = /(inputs|variables|assets|secrets)\.\w+/g;
```
bishal-pdMSFT, line 1: we also have `system`. Is that not being used in configuration mustache expressions?

src/configure/utilities/templateConverter.ts
```diff
@@ -0,0 +1,106 @@
...
+const templateEnclosedParametersRegex = /{{{?\s*(inputs|variables|assets|secrets)\.\w+\s*}?}}/g;
+const mustacheHelperRegex = /{?{{\s*#(\w+)(.*?)({?{{\s*\/\1)}}}?/;
```
bishal-pdMSFT, line 3: any reason why `{?` is applied to first `{` char and not third one like done one line above

bishal-pdMSFT, line 3: I could not understand this completely. Will catch you for this sometime.

src/configure/model/templateModels.ts
anuragc617: TemplateInfo is not being used anywhere. When we get the list of templateInfo, we directly store it in  RemotePipelineTemplate object and use it afterwards that's why workingDirectory is part of it.

src/configure/configure.ts
anuragc617: done

src/configure/configure.ts
anuragc617: done thanks for pointing it.

anuragc617: done.

src/configure/utilities/templateConverter.ts
anuragc617: I haven't seen it yet. I have added it in parameters regex.

src/configure/utilities/templateConverter.ts
anuragc617: `{?` means optional, it is there to match parameters even if they are 2 or 3 braces.

src/configure/clients/github/TemplateServiceClient.ts
anuragc617: done

src/configure/model/models.ts
anuragc617: it wasn't being used any where.

src/configure/utilities/templateConverter.ts
anuragc617: sure.

src/configure/configurers/configurerFactory.ts
anuragc617: yes, done

anuragc617:
> Please add proper telemetry for this flow.

already added telemetry.

PR #95, opened 2020-05-27
Issue fixing while flow testing

kanika1894:


src/configure/helper/mustacheHelper.ts
```diff
@@ -220,12 +220,14 @@ export class MustacheHelper {
         }
 
         var result: Object = {};
-        Object.keys(mustacheObject).forEach(key => {
-            if (!!key && !!mustacheObject[key]) {
-                result[key] = MustacheHelper.renderObject(mustacheObject[key], view);
-            }
-        });
+        if (mustacheObject) {
```
anuragc617, line 223: why this condition?

src/configure/templateInputHelper/InputControlProvider.ts
```diff
@@ -60,9 +60,12 @@ export class InputControlProvider {
             let dependentInputControlArray = this._getInputDependencyArray(inputControl, [properties[element]], false);
             let dependentClientInputMap = this._getClientDependencyMap(inputControl, [properties[element]]);
             let newValue = this._computeMustacheValue(properties[element], dependentInputControlArray, dependentClientInputMap);
+            if (!Object.keys(newValue).length) {
+                newValue = properties[element];
+            }
             inputControl.updateInputDescriptorProperty(key, newValue);
             if (key === constants.inputModeProperty) {
-                let updatedControlType = InputControlUtility.getInputControlType(InputMode[newValue]);
+                let updatedControlType = InputControlUtility.getInputControlType(parseInt(newValue.toString()));
```
anuragc617, line 68: why this change?

src/configure/templateInputHelper/InputControlProvider.ts
```diff
@@ -148,12 +151,11 @@ export class InputControlProvider {
     }
 
     private _computeMustacheValue(mustacheExpression: string, dependentInputControlArray: InputControl[], dependentClientInputMap: StringMap<any>): string {
-
         var dependentInputValues = this._getInputParameterValueIfAllSet(dependentInputControlArray);
         if (dependentInputControlArray && dependentInputControlArray.length > 0 && !dependentInputValues) {
             return "";
         } else {
-            return MustacheHelper.render(mustacheExpression, { inputs: dependentInputValues, client: dependentClientInputMap });
+            return MustacheHelper.renderObject(mustacheExpression, { inputs: dependentInputValues, client: dependentClientInputMap });
```
anuragc617, line 158: which case is not working with render()?

src/configure/configure.ts
```diff
@@ -171,6 +171,7 @@ class Orchestrator {
                 else {
                     this.context['rightClickScenario'] = true;
                 }
+                this.context['resourceId'] = this.inputs.targetResource.resource.id;
```
anuragc617, line 174: It will fail for non-webApp case ex AKS, because resource will be undefined there and it will throw an exception. 
Should we consider rightClickScenario=true for WebApp ??

src/configure/templateInputHelper/InputControlProvider.ts
kanika1894: Just because renderObject is also handling the string scenario and calling render for the same, so updated it just for improvement purpose.

src/configure/templateInputHelper/InputControlProvider.ts
kanika1894: The InputMode in template can be an integer or string. Example, if it's a override scenario, it will be treated as a string and direct inputMode property comes as a number. To handle both the scenarios, this is done.

src/configure/helper/mustacheHelper.ts
kanika1894: It throws a null exception if null is passed as a mustacheObject.
Example scenario : changing dataSourceId of a parameter to null in overrides property.

src/configure/configure.ts
kanika1894: ``
  private async getAzureResource(targetType: TargetResourceType) {
        var azureResourceSelector = ResourceSelectorFactory.getAzureResourceSelector(targetType);
    }
``

We are using azureResourceSelector to get the resourceId. In case of AKS, it will be null because the flow is different for now., but it doesn't seem this will cause a problem?

src/configure/templateInputHelper/InputControlProvider.ts
```diff
@@ -60,9 +60,12 @@ export class InputControlProvider {
             let dependentInputControlArray = this._getInputDependencyArray(inputControl, [properties[element]], false);
             let dependentClientInputMap = this._getClientDependencyMap(inputControl, [properties[element]]);
             let newValue = this._computeMustacheValue(properties[element], dependentInputControlArray, dependentClientInputMap);
+            if (!Object.keys(newValue).length) {
```
anuragc617, line 63: Why this condition is added?

PR #100, opened 2020-06-01
Adding separate node version converter for windows and linux

anuragc617:


src/configure/utilities/webAppNodeVersionConverter.ts
```diff
@@ -0,0 +1,36 @@
+import { JSONPath } from 'jsonpath-plus';
+import * as vscode from 'vscode';
+import { ArmRestClient } from '../clients/azure/armRestClient';
+import { AzureSession } from "../model/models";
+import { Messages } from '../resources/messages';
+
+export async function webAppRuntimeNodeVersionConverter(nodeVersion: string, armUri: string, azureSession: AzureSession): Promise<string> {
+    if (nodeVersion.indexOf('|') >= 0){
+        nodeVersion = nodeVersion.split('|')[1];
```
bishal-pdMSFT, line 9: If the first part is not node, then we should always convert to latest lts

src/configure/utilities/webAppNodeVersionConverter.ts
```diff
@@ -0,0 +1,36 @@
...
+                location: vscode.ProgressLocation.Notification,
+                title: Messages.GettingNodeVersion
+            },
+            async () => {
+                const resultSelector = "$.value[?(@.name === 'node')].properties.majorVersions[*].runtimeVersions";
+                const response = await new ArmRestClient(this.azureSession).fetchArmData(armUri, 'GET');
+                versions = JSONPath({ path: resultSelector, json: response, wrap: false, flatten: true });
+            }
+        );
+        let maxVersion = 0;
```
bishal-pdMSFT, line 24: Initialise to current lts version

src/configure/utilities/webAppNodeVersionConverter.ts
```diff
@@ -0,0 +1,36 @@
...
+        versions.forEach((version: string) => {
+            const match = version.match(/(\d+)-lts/);
+            if (match && match.length > 1) {
+                maxVersion = Math.max(maxVersion, +match[1]);
```
bishal-pdMSFT, line 28: Nice 😄

bishal-pdMSFT, line 28: Should it be match[0] ? I am not sure

anuragc617: No, groups are stored from 1st index onwards. 0th contains the regex match.

src/configure/utilities/webAppNodeVersionConverter.ts
anuragc617: If first part is not node, can the nodeVersion be any standard version? 
I am returning nodeVersion as it is if that is the case.

src/configure/utilities/webAppNodeVersionConverter.ts
anuragc617: Here we are comparing the major version which is numeric starts with 0. Trying to find the lts version by selecting the maximum major version.

src/configure/utilities/webAppNodeVersionConverter.ts
anuragc617: thanks..

src/configure/utilities/webAppNodeVersionConverter.ts
bishal-pdMSFT: If first part is not node, then it will lostly be not valid node version . Hence I thought we can do lts

src/configure/utilities/webAppNodeVersionConverter.ts
bishal-pdMSFT: Got it. I just wanted to handle the case when no match happens or an error ocuurs. In such case we can fall back to current lts.

src/configure/utilities/webAppNodeVersionConverter.ts
anuragc617: okk done

src/configure/utilities/webAppNodeVersionConverter.ts
anuragc617: falling back to lts will result in setup task failure.

PR #109, opened 2020-06-12
Added creation of GitHub repository feature

JA-04:
The following document describes the updated workflow.
[Setting up GitHub Repo.docx](https://github.com/microsoft/vscode-deploy-azure/files/4770276/Setting.up.GitHub.Repo.docx)


src/configure/clients/azure/appServiceClient.ts
```diff
@@ -48,9 +48,9 @@ export class AppServiceClient extends AzureResourceClient {
     public async getWebAppPublishProfileXml(resourceId: string): Promise<string> {
         let parsedResourceId: ParsedAzureResourceId = new ParsedAzureResourceId(resourceId);
         let publishingProfileStream = await this.webSiteManagementClient.webApps.listPublishingProfileXmlWithSecrets(parsedResourceId.resourceGroup, parsedResourceId.resourceName, {});
-        while (!publishingProfileStream.readable) {
+        /*while (!publishingProfileStream.readable) {
             // wait for stream to be readable.
-        }
+        }*/
```
kanika1894, line 53: This shouldn't be commented. It was only for understanding the code flow.

src/configure/clients/github/githubClient.ts
```diff
@@ -1,66 +1,143 @@
...
-        let sodiumObj = new SodiumLibHelper(secretKeyObject.key);
-        let encryptedBytes: Uint8Array = sodiumObj.encrypt(body);
-        let encryptedBytesAsString: string = SodiumLibHelper.convertUint8ArrayToString(encryptedBytes);
-        let encryptedEncodedText = SodiumLibHelper.encodeToBase64(encryptedBytesAsString);
-        await this._setGithubSecret(secretName, secretKeyObject.key_id, encryptedEncodedText);
+    public createGithubRepo(orgName: string, repoName: string): Promise<void | GitHubRepo> {
+        let restClient =new RestClient();
+        try{
+            return restClient.sendRequest(<UrlBasedRequestPrepareOptions>{
+                url: "https://api.github.com/orgs/" + orgName + "/repos",
```
kanika1894, line 28: Declare the fixed url, I guess this : https://api.github.com/orgs/ as a constant and reuse it at the places required.

src/configure/clients/github/githubClient.ts
```diff
@@ -1,66 +1,143 @@
...
+                    organizations = organizations.sort((org1, org2) => stringCompareFunction(org1.login, org2.login));
+                    return organizations;
+                });
+        }
+        return this.listOrgPromise; 
     }
 
-    private async _setGithubSecret(secretName: string, key_id: string, encrypted_secret: string): Promise<void> {
+    public async generateGitHubRepository(orgName: string, localPath: string): Promise<GitHubRepo | void>{
+        let repoName = localPath.substring(localPath.lastIndexOf("\\")+1).replace("*","-"); //need to include all the special characters here
```
kanika1894, line 82: Can you please explain why are we replacing the special characters?

src/configure/clients/github/githubClient.ts
```diff
@@ -1,66 +1,143 @@
...
+        let repoDetails = await this.createGithubRepo(orgName, repoName) as unknown as GitHubRepo | void;
+        if(repoDetails){
+            return repoDetails; 
+        }
+        repoName = repoName+"_"+orgName;
```
kanika1894, line 87: You can put this in else and put one liner comment for each case, related to what is the difference in the 3 cases

src/configure/clients/azure/appServiceClient.ts
JA-04: Okay, done.

src/configure/clients/github/githubClient.ts
```diff
@@ -1,66 +1,143 @@
...
+        }
+        catch(error){
+            return null;
+        }
     }
 
-    private async _getGitHubSecretKey() : Promise<GitHubSecretKey> {
-        let request = <UrlBasedRequestPrepareOptions> {
-                url: GitHubProvider.getFormattedGitHubApiUrlBase(this.url) + "/actions/secrets/public-key",
+    public async listOrganizations(PATtoken: string, forceRefresh?: boolean): Promise<void | GitHubOrganization[]> {
```
kanika1894, line 58: Move all the public functions above, followed by all private functions

src/configure/configurers/localGithubWorkflowConfigurer.ts
```diff
@@ -9,38 +9,83 @@ import { UserCancelledError } from 'vscode-azureextensionui';
...
+            throw error;
+        }
+        
     }
 
     public async validatePermissions(): Promise<void> {
         return;
     }
 
+
```
kanika1894, line 88: Remove the extra line. Just keep one blank line after end of each function.

src/configure/resources/messages.ts
```diff
@@ -102,4 +102,8 @@ export class Messages {
     public static gettingTemplateFileAsset = "Getting template file asset to commit";
     public static gettingWorkflowFile = "Getting workflow file";
     public static templateFileNotFound = "Template file %s not found";
+    public static selectGitHubOrganizationName = "Select a GitHUb organization";
```
kanika1894, line 105: Github*

src/configure/resources/messages.ts
```diff
@@ -102,4 +102,8 @@ export class Messages {
...
+    public static createGitHubOrganization = "Create a GitHub Organization first to create GitHub repository.";
+    public static newGitHubRepositoryCreated = "New GitHub repository with the name '%s' has been created.";
+    public static cannotCreateGitHubRepository = "New GitHub repository could not be created as the repository with the same already exists";
```
kanika1894, line 108: with the same name*

tslint.json
```diff
@@ -49,6 +49,6 @@
 			}
 		]
 	},
-	"defaultSeverity": "warning",
-	"extends": ["tslint-microsoft-contrib/recommended", "tslint:recommended"]
+	"defaultSeverity": "warning"
+	//"extends": ["tslint-microsoft-contrib/recommended", "tslint:recommended"]
```
kanika1894: Remove this comment as well

JA-04: Okay

src/configure/clients/github/githubClient.ts
JA-04: Putting it in 'else' will give rise to nested if-else which will lengthen the code, that's why I have avoided the use of 'else'. Will add the comments for the 3 cases to make the code more understandable.

package.json
```diff
@@ -108,15 +108,20 @@
         "vscode-azureextensiondev": "^0.3.1"
     },
     "dependencies": {
+        "@octokit/rest": "^17.11.0",
         "azure-arm-resource": "7.3.0",
```
bishal-pdMSFT, line 112: Lots of new packages being added? Are all of these used? I do not see them being imported anywhere

src/configure/clients/github/githubClient.ts
```diff
@@ -1,69 +1,150 @@
...
+                url: Url,
+                headers: {
+                    "Content-Type": "application/json",
+                    "Authorization": "Bearer " + this.patToken,
+                    "User-Agent": UserAgent
+                },
+                method: 'POST',
+                body: {
+                    "name": repoName,
+                    "description": "Repo created from VScode",
```
bishal-pdMSFT, line 38: should the description be `Repo created from VScode extension "Deploy to Azure"`

src/configure/clients/github/githubClient.ts
```diff
@@ -1,69 +1,150 @@
...
     }
 
-    public async createOrUpdateGithubSecret(secretName: string, body: string): Promise<void> {
-        let secretKeyObject: GitHubSecretKey = await this._getGitHubSecretKey();
-        let sodiumObj = new SodiumLibHelper(secretKeyObject.key);
-        let encryptedBytes: Uint8Array = sodiumObj.encrypt(body);
-        let encryptedBytesAsString: string = SodiumLibHelper.convertUint8ArrayToString(encryptedBytes);
-        let encryptedEncodedText = SodiumLibHelper.encodeToBase64(encryptedBytesAsString);
-        await this._setGithubSecret(secretName, secretKeyObject.key_id, encryptedEncodedText);
+    public createGithubRepo(orgName: string, repoName: string): Promise<void | GitHubRepo> {
```
bishal-pdMSFT, line 24: why `void | GitHubRepo` and not just `GitHubRepo`?

src/configure/clients/github/githubClient.ts
```diff
@@ -1,69 +1,150 @@
...
+        let restClient =new RestClient();
+        let Url = "https://api.github.com/orgs/" + orgName + "/repos";
+        try{
```
bishal-pdMSFT, line 27: do we need a try-catch here as the exception would anyways be handled by caller?

src/configure/clients/github/githubClient.ts
```diff
@@ -1,69 +1,150 @@
...
+                    "homepage": "https://github.com",
+                    "private": false,
+                    "has_issues": true,
+                    "has_projects": true,
+                    "has_wiki": true
+                },
+                deserializationMapper: null,
+                serializationMapper: null
+            })
+                .then((detail: GitHubRepo) => {
```
bishal-pdMSFT, line 48: you don't need the `then` and `catch` blocks as both success and failure will anyways propagate to caller

src/configure/clients/github/githubClient.ts
```diff
@@ -1,69 +1,150 @@
...
+                    return detail;
+                }).catch(error=>{
+                        return null;
```
bishal-pdMSFT, line 51: why do you want to suppress the error and return null instead?

src/configure/clients/github/githubClient.ts
```diff
@@ -1,69 +1,150 @@
...
+        }
+        catch(error){
+            return null;
+        }
     }
 
-    private async _getGitHubSecretKey() : Promise<GitHubSecretKey> {
-        let request = <UrlBasedRequestPrepareOptions> {
-                url: GitHubProvider.getFormattedGitHubApiUrlBase(this.url) + "/actions/secrets/public-key",
+    public async listOrganizations(PATtoken: string, forceRefresh?: boolean): Promise<void | GitHubOrganization[]> {
```
bishal-pdMSFT, line 59: PAT token has already been set in the ctor. It should not be needed here.

src/configure/clients/github/githubClient.ts
```diff
@@ -1,69 +1,150 @@
...
-                serializationMapper: null,
-                deserializationMapper: null
-            }
-        let restClient = new RestClient();
-        return (await restClient.sendRequest(request)) as GitHubSecretKey;
+                deserializationMapper: null,
+                serializationMapper: null
+            })
+                .then((organizations: Array<GitHubOrganization>) => {
+                    organizations = organizations.sort((org1, org2) => stringCompareFunction(org1.login, org2.login));
```
bishal-pdMSFT, line 75: ideally sorting should be done by caller as different caller might want an unsorted list

src/configure/clients/github/githubClient.ts
```diff
@@ -1,69 +1,150 @@
...
+                    return organizations;
+                });
+        }
+        return this.listOrgPromise; 
     }
 
-    private async _setGithubSecret(secretName: string, key_id: string, encrypted_secret: string): Promise<void> {
+    public async generateGitHubRepository(orgName: string, localPath: string): Promise<GitHubRepo | void>{
```
bishal-pdMSFT, line 82: the client class is strictly only for making rest calls. for wrapper functionalities on top of it, you should use existing `gitHubHelper` class in src\configure\helper\gitHubHelper.ts

package.json
JA-04: Earlier I was using Octokit to push files into GitHub repository, this was using globby objects and some functions from the package fs-extra, but since now simple-git is being used to push the files, these packages are rendered useless. 
Similarly, earlier we were taking the repo name as input from users, that's why package github-reserved-names was installed.
I'll delete all these packages.

src/configure/clients/github/githubClient.ts
bishal-pdMSFT: or you put it in commonHelper

src/configure/clients/github/githubClient.ts
```diff
@@ -1,69 +1,150 @@
...
+        let repoName = localPath.substring(localPath.lastIndexOf("\\")+1).replace("*","-"); //need to include all the special characters here
```
bishal-pdMSFT, line 83: why do you want to include special chars? why not remove all non-alphanumeric characters?

src/configure/clients/github/githubClient.ts
```diff
@@ -1,69 +1,150 @@
...
+        let repoDetails = await this.createGithubRepo(orgName, repoName) as unknown as GitHubRepo | void;
```
bishal-pdMSFT, line 84: do we need these type casts? Try to remove them if not really needed

src/configure/clients/github/githubClient.ts
JA-04: Since it's an API call, due to some error (like existing repo name) the function might not return a GitHubRepo object. Therefore, the return type 'void' was needed to be added. Also, by using 'void', it becomes convenient to move to other cases, where organization name or uuid needs to be appended in order to give a unique name to the GitHub repository.

src/configure/clients/github/githubClient.ts
```diff
@@ -1,69 +1,150 @@
...
+            return repoDetails; 
+        }
+        //Case: Local repo name is not available, therefore organization name has been appended
+        repoName = repoName+"_"+orgName;
+        repoDetails = await this.createGithubRepo(orgName, repoName) as unknown as GitHubRepo | void;
+        if(repoDetails){
+            return repoDetails; 
+        }
+        //Case: If the above two names are not available, uuid is also appended
+        return await this.createGithubRepo(orgName, repoName+"-"+uuid().substr(0,5));   //need to add proper Uiid or goid
```
bishal-pdMSFT, line 96: you can put spaces to increase readability. It is not bash :smile:

src/configure/clients/github/githubClient.ts
```diff
@@ -1,69 +1,150 @@
...
+    }
+
+    public async createOrUpdateGithubSecret(secretName: string, body: string): Promise < void> {
```
bishal-pdMSFT, line 99: can you please move this function out to the helper class as well.

src/configure/clients/github/githubClient.ts
```diff
@@ -1,69 +1,150 @@
...
+    let secretKeyObject: GitHubSecretKey = await this._getGitHubSecretKey();
+    let sodiumObj = new SodiumLibHelper(secretKeyObject.key);
+    let encryptedBytes: Uint8Array = sodiumObj.encrypt(body);
+    let encryptedBytesAsString: string = SodiumLibHelper.convertUint8ArrayToString(encryptedBytes);
+    let encryptedEncodedText = SodiumLibHelper.encodeToBase64(encryptedBytesAsString);
+    await this._setGithubSecret(secretName, secretKeyObject.key_id, encryptedEncodedText);
+}
+
+    private async _getGitHubSecretKey() : Promise < GitHubSecretKey > {
```
bishal-pdMSFT, line 108: make this public

src/configure/clients/github/githubClient.ts
```diff
@@ -1,69 +1,150 @@
...
-            deserializationMapper: null,
-            serializationMapper: null,
-            body: {
-                "encrypted_value": encrypted_secret,
-                "key_id": key_id
-            }
+    return(await restClient.sendRequest(request)) as GitHubSecretKey;
+}
+
+    private async _setGithubSecret(secretName: string, key_id: string, encrypted_secret: string): Promise < void> {
```
bishal-pdMSFT, line 125: make this public

src/configure/clients/github/githubClient.ts
JA-04: Will change it.

src/configure/clients/github/githubClient.ts
JA-04: The caller is not handling the exception, it is just deciding which case to execute (out of the 3 cases for naming GitHub repo name) depending upon the return type of this function. By doing so, the code is shortened as we don't need nested try-catch blocks.

src/configure/clients/github/githubClient.ts
JA-04: Because there are considerable chances of occurrence of an error (already existing repo of the same name),  therefore the error is suppressed here and void is returned. On receiving 'void', caller function will know that the repo name already exists and therefore will append organization name or uuid to make the repo name unique.

src/configure/clients/github/githubClient.ts
JA-04: The sorting feature was included here because sorting is done in case of Azure pipeline as well. Should I remove it?

src/configure/configurers/localGithubWorkflowConfigurer.ts
```diff
@@ -9,32 +9,75 @@ import { UserCancelledError } from 'vscode-azureextensionui';
...
+                    if(newGitHubRepo){
+                        inputs.sourceRepository.remoteName = newGitHubRepo.name;
+                        inputs.sourceRepository.remoteUrl = newGitHubRepo.html_url+".git";
+                        inputs.sourceRepository.repositoryId = GitHubProvider.getRepositoryIdFromUrl(inputs.sourceRepository.remoteUrl);
+                        await this.localGitRepoHelper.initializeGitRepository(inputs.sourceRepository.remoteName, inputs.sourceRepository.remoteUrl);
+                        vscode.window.showInformationMessage(utils.format(Messages.newGitHubRepositoryCreated, newGitHubRepo.name));
+                    }
+                    else{
+                        vscode.window.showErrorMessage(Messages.cannotCreateGitHubRepository);
+                        throw Error;
```
bishal-pdMSFT, line 67: throwing empty error is not very good as it will not help while analyzing the failures. You can throw same error messages as what you are showing to user.

src/configure/clients/github/githubClient.ts
JA-04: Without these typecasts the editor is giving errors.

src/configure/configurers/localGithubWorkflowConfigurer.ts
```diff
@@ -9,32 +9,75 @@ import { UserCancelledError } from 'vscode-azureextensionui';
...
+                    
+                }
+                else{
+                    vscode.window.showErrorMessage(Messages.createGitHubOrganization);
+                    throw Error;
+                }
+            }
+        }
+        catch (error) {
+            telemetryHelper.logError(Layer, TracePoints.GetGitHubDetailsFailed, error);
```
bishal-pdMSFT, line 78: you have already logged two useful error telemetry above. Do you really want to catch then again? If the chanced of other errors happening are very low, then remove this catch block as it will lead to double error telemetries being logged for same error.

src/configure/configurers/localGithubWorkflowConfigurer.ts
JA-04: Okay, will change it.

src/configure/configurers/localGithubWorkflowConfigurer.ts
bishal-pdMSFT: Ohh sorry, above we are only showing error and not logging telemetry. It should be fine then.
But do think about this. I think it would be better to log telemetry there only as they are very pointed telemetry point unlike GetGitHubDetailsFailed which is very generic. We can remove this catch block in such a case

src/configure/model/models.ts
```diff
@@ -52,6 +52,29 @@ export class Organization {
     isMSAOrg: boolean;
 }
 
+export class GitHubOrganization {
+    login: string;
```
bishal-pdMSFT, line 56: are all these properties mandatory for GitHub API and are we using all? If no, then just keep such fields.

src/configure/model/models.ts
```diff
@@ -52,6 +52,29 @@ export class Organization {
...
+    public_members_url: string;
+    avatar_url: string;
+    description: string;
+}
+
+export class GitHubRepo{
+    id: string;
+    name: string;
+    orgName: string;
+    html_url: string;
```
bishal-pdMSFT, line 75: is this required?

src/configure/configure.ts
```diff
@@ -356,7 +356,7 @@ class Orchestrator {
             remoteUrl: '',
             repositoryId: '',
             repositoryName: '',
-            repositoryProvider: RepositoryProvider.AzureRepos
+            repositoryProvider: RepositoryProvider.Github,
```
bishal-pdMSFT, line 359: This should be controlled via a VS Code [setting](https://code.visualstudio.com/docs/getstarted/settings). That setting will allow user to switch back to Azure repo if they want. This also provides us a fallback if customers start hitting issue with new flow of GitHub repo.

bishal-pdMSFT: See one such setting here https://github.com/microsoft/vscode-deploy-azure/blob/c757820f632a2c4fc02fccefe2204dd9f29470aa/src/configure/model/models.ts#L21

src/configure/model/models.ts
JA-04: Yes, for receiving data from API call used for creating a new GitHub repository.

src/configure/model/models.ts
JA-04: No, we are not using all the fields. I'll remove the unused fields.

src/configure/clients/github/githubClient.ts
JA-04: The caller is not handling the exception that's why 'then' and 'catch' blocks are used to handle failure and success cases separately.

src/configure/clients/github/githubClient.ts
JA-04: I have removed the 'replace' function now since GitHub automatically handles the special cases.

src/configure/clients/github/githubClient.ts
JA-04: Okay, will rectify it.

src/configure/clients/github/githubClient.ts
JA-04: Will move the function to gitHubHelper.ts

src/configure/clients/github/githubClient.ts
JA-04: Sure.

src/configure/clients/github/githubClient.ts
JA-04: Okay

src/configure/clients/github/githubClient.ts
JA-04: Okay

src/configure/clients/github/githubClient.ts
JA-04: Sure, will keep in mind from now on to make the code more readable :)

src/configure/configurers/localGithubWorkflowConfigurer.ts
JA-04: Yeah, logging the telemetry in the 'else' block will be a better option. And it will also eliminate the need of try-catch block. I'll make the required changes.

src/configure/configure.ts
JA-04: Okay, will add this setting.

src/configure/clients/github/githubClient.ts
JA-04: This part of code was added earlier but it's now removed as GitHub automatically handles the special characters in the repository name.

src/configure/configurers/remoteGitHubWorkflowConfigurer.ts
```diff
@@ -139,7 +140,7 @@ export class RemoteGitHubWorkflowConfigurer extends LocalGitHubWorkflowConfigure
                             title: Messages.settingUpGithubSecrets
                         },
                         async () => {
-                            await this.githubClient.createOrUpdateGithubSecret(secretKey, secretValue);
+                            await createOrUpdateGithubSecret(secretKey, secretValue);
```
vineetmimrot, line 143: You can convert the createOrUpdateGithubSecret() inside GitHub client to static, instead of defining it again.

src/configure/clients/github/githubClient.ts
vineetmimrot: Isn't the failure may also happen when user doesn't have permission to create repo in the current org?? So it is better to bump up the error to the user.

package.json
```diff
@@ -76,6 +76,11 @@
                     "type": "boolean",
                     "default": false,
                     "description": "Setup CI/CD in Azure Pipelines for Github Repository"
+                },
+                "deployToAzure.SetGitHubAsDefaultRepositoryProvider": {
+                    "type": "boolean",
+                    "default": true,
+                    "description": "Use GitHub as default repository provider"
```
bishal-pdMSFT, line 83: "default repository provider" seems like a very generic term. Should we call it "Use GitHub for creating new repository"?

package.json
```diff
@@ -114,9 +119,10 @@
         "js-yaml": "^3.13.1",
         "jsonpath-plus": "^3.0.0",
         "mustache": "3.0.1",
+        "npm": "^6.14.5",
```
bishal-pdMSFT, line 122: why do we need npm as a dependency?

src/configure/clients/github/githubClient.ts
```diff
@@ -9,61 +10,113 @@ export class GithubClient {
...
+                    "has_projects": true,
+                    "has_wiki": true
+                },
+                deserializationMapper: null,
+                serializationMapper: null
+            })
+                .then((detail: GitHubRepo) => {
+                    return detail;
+                }).catch(error=>{
+                        return null;
```
bishal-pdMSFT, line 49: please log the error here in telemetry

package.json
JA-04: Sure. Will change it.

src/configure/helper/commonHelper.ts
```diff
@@ -11,6 +16,32 @@ export async function sleepForMilliSeconds(timeInMs: number): Promise<void> {
...
+        return repoDetails; 
+    }
+    //Case: Local repository name is not available, therefore organization name has been appended
+    repoName = repoName + "_" + orgName;
+    repoDetails = await githubClient.createGithubRepo(orgName, repoName) as GitHubRepo | void;
+    if(repoDetails){
+        return repoDetails; 
+    }
+    //Case: If the above two repository names are not available, uuid is also appended
+    return await githubClient.createGithubRepo(orgName, repoName + "-" + uuid().substr(0,5));
```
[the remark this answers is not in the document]
bishal-pdMSFT: it would be weird to stop the flow without telling user how they can fix it. Either we ask the repo name from user or append a unique guid. Unique guid seems to be simpler approach. And the name will still be helpful as it starts with org+repo name

src/configure/clients/github/githubClient.ts
bishal-pdMSFT: Okay. I think you need not specify `then` as it will be bubbled up to caller automatically. You can try.

src/configure/clients/github/githubClient.ts
bishal-pdMSFT: That is correct. I think you should pass a variable `shouldThrow` as argument. If true, then do not suppress and throw. caller will set `false` first two times but will set `true` the last time.

src/configure/clients/github/githubClient.ts
```diff
@@ -9,61 +10,113 @@ export class GithubClient {
...
-                serializationMapper: null,
-                deserializationMapper: null
-            }
-        let restClient = new RestClient();
-        return (await restClient.sendRequest(request)) as GitHubSecretKey;
+                deserializationMapper: null,
+                serializationMapper: null
+            })
+                .then((organizations: Array<GitHubOrganization>) => {
+                    organizations = organizations.sort((org1, org2) => stringCompareFunction(org1.login, org2.login));
```
bishal-pdMSFT, line 73: Is there a way to get only those orgs where user has create repo permission?

src/configure/clients/github/githubClient.ts
JA-04: Will do.

src/configure/clients/github/githubClient.ts
JA-04: No, none of the GitHub APIs provides this feature.

src/configure/clients/github/githubClient.ts
JA-04: @bishal-pdMSFT  If repository cannot be created, then localGithubWorkflowConfigurer.ts is already throwing an error. Should I just change the error message to "GitHub repository cannot be created. You either don't have permission to create a repository inside the selected organization or a repository with the same name already exists."

package.json
JA-04: It was added automatically, will remove it.

src/configure/clients/github/githubClient.ts
JA-04: Without `then`, the errors are not getting caught properly.

PR #110, opened 2020-06-15
Use event to read publish profile from stream instead of sync way

bishal-pdMSFT:
This fixes the bug where reading publish profile hangs waiting for the stream to become readable.

The root cause of the hang was the fact that ms-rest library mandatorily JSON-deserializes the ARM response. In case of publish profile, it was failing to de-serialize as publish profiles contain double quotes. The fix to handle this is to escape the quotes and then JSON stringify.

Added a filter which is a way provided by ms-rest to add filters in the request pipeline. First tried to add it to AppServiceClient itself but somehow azure-arm-website library was not handling the filter or the output properly. Since, there is less level of control over the website client as it is part of azure-arm-website package, I decided to use a lower level library - ArmRestClient. This client has been used for AKS related calls, so it seems fine to use for websites as well.


bishal-pdMSFT:
@kanika1894 @vineetmimrot @anuragc617 please review

bishal-pdMSFT:
Abandoning as Anurag had a better fix.

PR #111, opened 2020-06-16
Fixing issues - vscode bugbash169

kanika1894:
1. AKS template issue : https://mseng.visualstudio.com/AzureDevOps/_workitems/edit/1733192/
2. dotnet core application - azure pipeline build issue : 
https://mseng.visualstudio.com/AzureDevOps/_workitems/edit/1733196/
3. handling for not supported language : https://mseng.visualstudio.com/AzureDevOps/_workitems/edit/1733189/

### DIFF
--- a/src/configure/configure.ts
+++ b/src/configure/configure.ts
@@ -493,6 +493,10 @@ class Orchestrator {
         }
         let pipelineMap = this.getMapOfUniqueLabels(appropriatePipelines);
         let pipelineLabels = Array.from(pipelineMap.keys());
+        if (pipelineLabels.length === 0) {
+            telemetryHelper.setTelemetry(TelemetryKeys.UnsupportedLanguage, Messages.languageNotSupported);
+            throw new Error(Messages.languageNotSupported);
+        }
 
         // TO:DO- Get applicable pipelines for the repo type and azure target type if target already selected
         if (pipelineLabels.length > 1) {
--- a/src/configure/configure.ts
+++ b/src/configure/configure.ts
@@ -147,7 +147,7 @@ class Orchestrator {
                     this.inputs.pipelineConfiguration.template = shortlistedTemplates[0];
                 }
                 else {
-                    telemetryHelper.logError(Layer, TracePoints.SelectTemplate, new Error(Messages.TemplateNotFound));
+                    telemetryHelper.logError(Layer, TracePoints.TemplateNotFound, new Error(Messages.TemplateNotFound + " RepoId: " + this.inputs.sourceRepository.repositoryId));
                     throw new Error(Messages.TemplateNotFound);
                 }
                 break;
--- a/src/configure/configure.ts
+++ b/src/configure/configure.ts
@@ -128,6 +128,8 @@ class Orchestrator {
     private async getAzureResource(targetType: TargetResourceType) {
         var azureResourceSelector = ResourceSelectorFactory.getAzureResourceSelector(targetType);
         this.inputs.targetResource.resource = await azureResourceSelector.getAzureResource(this.inputs);
+        telemetryHelper.setTelemetry(TelemetryKeys.resourceType, this.inputs.targetResource.resource.type);
+        telemetryHelper.setTelemetry(TelemetryKeys.resourceKind, this.inputs.targetResource.resource.kind);
     }
 
     private async selectTemplate(resource: GenericResource): Promise<void> {
@@ -145,6 +147,7 @@ class Orchestrator {
                     this.inputs.pipelineConfiguration.template = shortlistedTemplates[0];
                 }
                 else {
+                    telemetryHelper.setTelemetry(TelemetryKeys.TemplateNotFound, Messages.TemplateNotFound);
                     throw new Error(Messages.TemplateNotFound);
                 }
                 break;
@@ -171,10 +174,13 @@ class Orchestrator {
                 else {
                     this.context['rightClickScenario'] = true;
                 }
+
                 if (this.inputs.targetResource.resource && this.inputs.targetResource.resource.id) {
                     this.context['resourceId'] = this.inputs.targetResource.resource.id;
                 }
                 this.selectTemplate(this.inputs.targetResource.resource);
+                telemetryHelper.setTelemetry(TelemetryKeys.SelectedTemplate, this.inputs.pipelineConfiguration.template.label);
+                telemetryHelper.setTelemetry(TelemetryKeys.SelectedTemplateType, (this.inputs.pipelineConfiguration.template.templateType).toString());
 
                 await this.updateRepositoryAnalysisApplicationSettings(repoAnalysisResult);
 
--- a/src/configure/configure.ts
+++ b/src/configure/configure.ts
@@ -147,7 +147,7 @@ class Orchestrator {
                     this.inputs.pipelineConfiguration.template = shortlistedTemplates[0];
                 }
                 else {
-                    telemetryHelper.setTelemetry(TelemetryKeys.TemplateNotFound, Messages.TemplateNotFound);
+                    telemetryHelper.logError(Layer, TracePoints.SelectTemplate, new Error(Messages.TemplateNotFound));
                     throw new Error(Messages.TemplateNotFound);
                 }
                 break;
--- a/src/configure/helper/AssetHandler.ts
+++ b/src/configure/helper/AssetHandler.ts
@@ -163,7 +163,7 @@ export class AssetHandler {
      */
     public static getSanitizedUniqueAssetName(assetName: string): string {
         assetName = assetName + "_" + UniqueResourceNameSuffix;
-        assetName = assetName.replace(/\W/, '');
+        assetName = assetName.replace(/\W/g, '');
         return assetName;
     }
 }
--- a/src/configure/resources/tracePoints.ts
+++ b/src/configure/resources/tracePoints.ts
@@ -28,4 +28,5 @@ export class TracePoints {
     public static GetRepoAnalysis = 'GetRepoAnalysis';
     public static GetTemplateParameters = 'GetTemplateParameters';
     public static GetTemplateConfiguration = 'getTemplateConfiguration';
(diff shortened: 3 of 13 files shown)


bishal-pdMSFT:
@kanika1894  can you please link the mseng bug here for easy reference

bishal-pdMSFT:
Also, @kanika1894 can you please highlight the changes in the created workflow which fixed the run?

src/configure/helper/AssetHandler.ts
```diff
@@ -163,7 +163,7 @@ export class AssetHandler {
      */
     public static getSanitizedUniqueAssetName(assetName: string): string {
         assetName = assetName + "_" + UniqueResourceNameSuffix;
-        assetName = assetName.replace(/\W/, '');
+        assetName = assetName.replace(/\W/g, '');
```
kanika1894, line 166: https://mseng.visualstudio.com/AzureDevOps/_workitems/edit/1733192/

src/configure/templates/azurePipelineTemplates/dotnetcoreLinuxWebApp.yml
```diff
@@ -23,10 +23,10 @@ variables:
   buildConfiguration: 'Release'
 
   # Build Projects
-  buildProjects: "$(workingDirectory)/**/*.csproj"
+  buildProjects: "{{#equals}}{{{ workingDirectory }}} '.' true '**/*.csproj' '$(workingDirectory)/**/*.csproj'{{/equals}}"
```
kanika1894, line 26: https://mseng.visualstudio.com/AzureDevOps/_workitems/edit/1733196/

src/configure/configure.ts
```diff
@@ -493,6 +499,10 @@ class Orchestrator {
         }
         let pipelineMap = this.getMapOfUniqueLabels(appropriatePipelines);
         let pipelineLabels = Array.from(pipelineMap.keys());
+        if (pipelineLabels.length === 0) {
+            telemetryHelper.setTelemetry(TelemetryKeys.UnsupportedLanguage, Messages.languageNotSupported);
+            throw new Error(Messages.languageNotSupported);
+        }
```
kanika1894, line 505: https://mseng.visualstudio.com/AzureDevOps/_workitems/edit/1733189/

src/configure/templateInputHelper/RepoAnalysisSettingInputProvider.ts
```diff
@@ -25,13 +25,18 @@ export class RepoAnalysisSettingInputProvider {
...
-                let error = new Error(`RepostioryAnalysisSetting doesn't contain ${repoAnalysisSettingKey} for input ${inputControl.getInputControlId()}`);
-                telemetryHelper.logError(Layer, TracePoints.SetInputControlValueFromRepoAnalysisResult, error);
-                let value = await new ControlProvider().showInputBox(repoAnalysisSettingKey, {
-                    placeHolder: inputControl.getInputDescriptor().name,
-                    validateInput: value => inputControl.triggerControlValueValidations(value)
-                });
-                inputControl.setValue(value);
+                if (inputControl.getInputDescriptor().defaultValue) {
+                    let value = inputControl.getInputDescriptor().defaultValue;
+                    inputControl.setValue(value);
```
kanika1894, line 30: This will pickup default value of an input if repoAnalysisSetting key doesn't exist. It handles webpack.config scenario.

src/configure/configure.ts
```diff
@@ -128,6 +128,8 @@ class Orchestrator {
     private async getAzureResource(targetType: TargetResourceType) {
         var azureResourceSelector = ResourceSelectorFactory.getAzureResourceSelector(targetType);
         this.inputs.targetResource.resource = await azureResourceSelector.getAzureResource(this.inputs);
+        telemetryHelper.setTelemetry(TelemetryKeys.resourceType, this.inputs.targetResource.resource.type);
```
bishal-pdMSFT, line 131: weren't these being already logged in some or other form?

src/configure/configure.ts
```diff
@@ -171,10 +174,13 @@ class Orchestrator {
                 else {
                     this.context['rightClickScenario'] = true;
                 }
+
                 if (this.inputs.targetResource.resource && this.inputs.targetResource.resource.id) {
                     this.context['resourceId'] = this.inputs.targetResource.resource.id;
                 }
                 this.selectTemplate(this.inputs.targetResource.resource);
+                telemetryHelper.setTelemetry(TelemetryKeys.SelectedTemplate, this.inputs.pipelineConfiguration.template.label);
```
bishal-pdMSFT, line 182: would it be better to log template Id instead of label?

src/configure/configure.ts
bishal-pdMSFT: Rather than failing here, we should use the simple application template.

bishal-pdMSFT: and that should happen for both azure pipelines and github worklow

src/configure/configure.ts
```diff
@@ -145,6 +147,7 @@ class Orchestrator {
                     this.inputs.pipelineConfiguration.template = shortlistedTemplates[0];
                 }
                 else {
+                    telemetryHelper.setTelemetry(TelemetryKeys.TemplateNotFound, Messages.TemplateNotFound);
```
bishal-pdMSFT, line 150: If I am not wrong `telemetryHelper.setTelemetry` will add this telemetry point to the command's log. That is good for any customer intelligence kind of data.
For exceptions, should we use `telemetryHelper.logError`

bishal-pdMSFT: The next line which throws the error will automatically add this error to the command's log (as `error` property) as well so no need add one extra property there.
If you want to log this error so that you can find all such template not found errors, you should use `telemetryHelper.logError`

src/configure/templateInputHelper/RepoAnalysisSettingInputProvider.ts
```diff
@@ -25,13 +25,18 @@ export class RepoAnalysisSettingInputProvider {
...
+                } else {
+                    let error = new Error(`RepostioryAnalysisSetting doesn't contain ${repoAnalysisSettingKey} for input ${inputControl.getInputControlId()}`);
+                    telemetryHelper.logError(Layer, TracePoints.SetInputControlValueFromRepoAnalysisResult, error);
```
bishal-pdMSFT, line 33: the error logging should happen irrespective of whether defaulvaule exists or not

src/configure/configure.ts
kanika1894: The local templates don't have template id.

src/configure/configure.ts
kanika1894: These are being logged in right click scenario, but were not there in the other flow. So added.

src/configure/templateInputHelper/RepoAnalysisSettingInputProvider.ts
kanika1894: updated

src/configure/configure.ts
kanika1894: This helps us in handling the existing bug. For scenario, where repo contains php and dockerFile, it will be an issue as we are supporting docker language but not having a template in azurePipeline. So, keeping this should be a good as it will handle the existing bug.

src/configure/configure.ts
kanika1894: updated

src/configure/configure.ts
```diff
@@ -145,6 +147,7 @@ class Orchestrator {
                     this.inputs.pipelineConfiguration.template = shortlistedTemplates[0];
                 }
                 else {
+                    telemetryHelper.logError(Layer, TracePoints.SelectTemplate, new Error(Messages.TemplateNotFound));
```
bishal-pdMSFT, line 150: Shouldn't the trace point be something like `TemplateNotFound`. This name does not sound very intuitive when querying for such failures.

bishal-pdMSFT: Also, should we add more details in the error message to tell us what is the scenario where we could not find any template. Should we log the repo id in the message. Or should we log the number of applicationSettings present in repo analysis result. These can help us diagnose what are the cases when no templates are being found.

bishal-pdMSFT: > Shouldn't the trace point be something like TemplateNotFound. This name does not sound very intuitive when querying for such failures.

Or is the idea to have a trace point for each action, and then use that trace points for all errors while executing that action. If so, this sounds fine.

src/configure/resources/tracePoints.ts
```diff
@@ -24,5 +24,9 @@ export class TracePoints {
     public static InitializeDynamicValidation = 'InitializeDynamicValidation';
     public static SetInputControlValueFromRepoAnalysisResult = 'setInputControlValueFromRepoAnalysisResult';
     public static TemplateServiceCallFailed = 'TemplateServiceCallFailed';
-    public static GetTemplateFile = 'GetTemplateFile'
+    public static GetTemplateFile = 'GetTemplateFile';
```
bishal-pdMSFT, line 27: It would be good if you can write a comment for each trace point - telling what kind of action errors will this trace point capture. This will help us in writing kusto queries when trying to figure our some specific falures.
Do it only for the new trace points. That should be okay.

src/configure/templates/azurePipelineTemplates/dotnetcoreLinuxWebApp.yml
bishal-pdMSFT: ahh yes, now I understand the issue. I could not get it from the build logs. Thanks :)

src/configure/configure.ts
kanika1894: yeah, another trace point will be a better idea. Also, will append repoId with the message.

src/configure/resources/tracePoints.ts
kanika1894: Updated the TracePoints for more clarity

PR #112, opened 2020-06-18
Adding publish profile asset

kanika1894:


src/configure/configurers/remoteGitHubWorkflowConfigurer.ts
```diff
@@ -169,6 +170,30 @@ export class RemoteGitHubWorkflowConfigurer extends LocalGitHubWorkflowConfigure
                         this.assets[asset.id] = await nodeVersionConverter.webAppRuntimeNodeVersionConverter(nodeVersion, armUri, this.azureSession);
                     }
                     break;
+                case "PublishProfile":
```
bishal-pdMSFT, line 173: this is good. But can you please remove SPN creation from pre-requisite to be part of asset creation. @anuragc617 is that going to cause problem?

anuragc617: yes we should move this from pre-requisite as otherwise it will create SPN for workflow. But any input to asset should be created before rendering it.

bishal-pdMSFT: in my mind, SPN is not an input necessarily. The auth input can be access token as well and the asset handler will create SPN as part of asset creation.

anuragc617: yes in ideal scenario the auth input can be: SPN, arm token or azureDevopAuth. How would we come to know that in vs code we need to create SPN, are we binding that to asset type? And if SPN is created as part of asset creation what is the use of auth input?

src/configure/configurers/remoteGitHubWorkflowConfigurer.ts
```diff
@@ -169,6 +170,30 @@ export class RemoteGitHubWorkflowConfigurer extends LocalGitHubWorkflowConfigure
...
+                                title: utils.format(Messages.creatingAzureServiceConnection, subscriptionId)
+                            },
+                            async () => {
+                                try {
+                                    // find LCS of all azure resource params
+                                    let appServiceClient = new AppServiceClient(this.azureSession.credentials, this.azureSession.environment, this.azureSession.tenantId, subscriptionId);
+                                    return await appServiceClient.getWebAppPublishProfileXml(resourceId);
+                                }
+                                catch (error) {
+                                    telemetryHelper.logError(Layer, TracePoints.AzureServiceConnectionCreateFailure, error);
```
bishal-pdMSFT, line 191: do you want to rename the tracepoint to GetAzurePublishProfile

kanika1894: updated

PR #115, opened 2020-06-28
Fix publish profile hanging issue using ServiceCallBack

anuragc617:


src/configure/clients/azure/appServiceClient.ts
```diff
@@ -46,24 +47,21 @@ export class AppServiceClient extends AzureResourceClient {
...
-        }
-
+        const deferred: Q.Deferred<string> = Q.defer<string>();
+        const parsedResourceId: ParsedAzureResourceId = new ParsedAzureResourceId(resourceId);
         let publishProfile = '';
-        while (true) {
-            let moreData: Buffer = publishingProfileStream.read();
-            if (moreData) {
-                publishProfile += moreData.toString();
+        this.webSiteManagementClient.webApps.listPublishingProfileXmlWithSecrets(parsedResourceId.resourceGroup, parsedResourceId.resourceName, { format: 'ftp' }, (err, result, request, response) => {
```
bishal-pdMSFT, line 53: why specific format?

src/configure/clients/azure/appServiceClient.ts
```diff
@@ -46,24 +47,21 @@ export class AppServiceClient extends AzureResourceClient {
...
+                throw err;
             }
-            else {
-                break;
-            }
-        }
-
-        return publishProfile;
+            response.on("data", (chunk: Buffer) => {
+                publishProfile += chunk;
```
bishal-pdMSFT, line 58: don't you want to do .toString()

src/configure/clients/azure/appServiceClient.ts
```diff
@@ -46,24 +47,21 @@ export class AppServiceClient extends AzureResourceClient {
...
+        const deferred: Q.Deferred<string> = Q.defer<string>();
+        const parsedResourceId: ParsedAzureResourceId = new ParsedAzureResourceId(resourceId);
         let publishProfile = '';
-        while (true) {
-            let moreData: Buffer = publishingProfileStream.read();
-            if (moreData) {
-                publishProfile += moreData.toString();
+        this.webSiteManagementClient.webApps.listPublishingProfileXmlWithSecrets(parsedResourceId.resourceGroup, parsedResourceId.resourceName, { format: 'ftp' }, (err, result, request, response) => {
+            if (err) {
+                throw err;
```
bishal-pdMSFT, line 55: should we do `deferred.reject` here as we already have deferred here

src/configure/clients/azure/appServiceClient.ts
anuragc617: yes it's not needed, removed it.

src/configure/clients/azure/appServiceClient.ts
anuragc617: It's not needed because it is auto-converted as publishProfile is string

src/configure/clients/azure/appServiceClient.ts
anuragc617: done

PR #116, opened 2020-06-29
Linting Support for GitHub workflow files

JA-04:
- Added linting support.
- Scope of linting is limited to GitHub workflow files only.
- Linting support includes the following features:
      1. Syntax checking
      2. Auto completion
      3. Hover Support
      4. Document outlining


syntaxes/yaml.tmLanguage.json
```diff
@@ -0,0 +1,621 @@
+{
```
kanika1894, line 1: What purpose does this file serve?

kanika1894:
A couple of binary files have been committed. Just verify what all needs to be included.

syntaxes/yaml.tmLanguage.json
JA-04: This file provides grammar which is used for syntax validation.

package.json
```diff
@@ -42,6 +42,29 @@
     ],
     "main": "./out/extension",
     "contributes": {
+        "languages": [
+            {
+                "id": "yaml",
+                "aliases": [
```
bishal-pdMSFT, line 48: what does aliases mean?

package.json
```diff
@@ -42,6 +42,29 @@
...
+                    "YAML",
+                    "yaml"
+                ],
+                "extensions": [
+                    ".yml",
+                    ".eyaml",
+                    ".eyml",
+                    ".yaml"
+                ],
+                "firstLine": "^#cloud-config"
```
bishal-pdMSFT, line 58: what does this mean?

JA-04: Initially, one of the ideas to limit linting scope was to use `firstLine`, but now since documentorSelector is solving the purpose it is not required. Will remove it.

package.json
JA-04: It just states the aliases of YAML language.

package.json
```diff
@@ -42,6 +42,28 @@
...
+                ],
+                "extensions": [
+                    ".yml",
+                    ".eyaml",
+                    ".eyml",
+                    ".yaml"
+                ]
+            }
+        ],
+        "grammars": [
```
bishal-pdMSFT, line 60: what is the purpose of grammar?

src/extension.ts
```diff
@@ -24,9 +25,45 @@ export async function activate(context: vscode.ExtensionContext) {
                     await activateConfigurePipeline();
             },
             TelemetryKeys.ExtensionActivationDuration);
-    });
+	});
+	
+    // The YAML language server is implemented in node
```
bishal-pdMSFT, line 30: Can this language client and server configuration fail? If so, we should move it inside try-catch and suppress (and send telemetry) any failure. This is an additional feature.

src/extension.ts
```diff
@@ -24,9 +25,45 @@ export async function activate(context: vscode.ExtensionContext) {
...
+		debug: { module: serverModule, transport: TransportKind.ipc, options: debugOptions }
+	};
+
+	// Options to control the scope of language client
+	let clientOptions: LanguageClientOptions = {
+		documentSelector: [
+			{language: "yaml", pattern: "**/.github/workflows/**", scheme: "file"}
+		],
+		synchronize: {
+			// Synchronize these setting sections with the server
```
bishal-pdMSFT, line 49: should we lock onto a fixed version of server? otherwise how will we know that we have diverged from server?

syntaxes/yaml.tmLanguage.json
bishal-pdMSFT: what syntax validation? yaml or workflow?

bishal-pdMSFT: looks like yaml syntax. so, the server does not provide this syntax validation?

JA-04: Thie syntax validation is for YAML files, in general, and not specifically for workflow files. The server provides only those features which are schema dependent like auto-completion and hover support.

package.json
JA-04: The grammar is used for syntax validation. Here, by providing the path to `yaml.tmLanguage.json`, we are specifying the grammar to be used for syntax validation.

src/extension.ts
JA-04: It might fail if some changes are made on the server-side. As a precaution, I'll add a try-catch block and log the telemetry.

bishal-pdMSFT: @JA-04 you can also move the code to a separate function just to keep activate() short and easy to read.

package.json
```diff
@@ -128,8 +150,10 @@
         "uuid": "^3.3.2",
         "vscode-azureextensionui": "0.26.3",
         "vscode-extension-telemetry": "0.0.18",
+        "vscode-languageclient": "^5.2.1",
         "vscode-nls": "3.2.4",
-        "vscode-uri": "1.0.6"
+        "vscode-uri": "1.0.6",
```
bishal-pdMSFT, line 155: where is this used?

src/extension.ts
bishal-pdMSFT: what I meant was whether we should use a fixed version instead of `^` in `"yaml-language-server": "^0.8.0"`

package.json
JA-04: It is used in `modaRepositoryAnalysisClient.ts`.

src/extension.ts
JA-04: Okay, will create a new function for this.

src/extension.ts
JA-04: It won't be required as `try-catch` block has already been added, in case our client becomes incompatible with server in future.

PR #124, opened 2020-07-02
Fixed repo creation bug

JA-04:
- Fixed the repo creation bug which occurred due to different path structure of different operating systems.
- Added error handling for repo creation inside organizations using SAML SSO.
- Added telemetry error logging for simple-git functions.
- Changed visibility of newly created repos to private.


src/configure/clients/restClient.ts
```diff
@@ -20,6 +20,21 @@ export class RestClient extends ServiceClient {
         });
     }
 
+    public sendRequest3<TResult>(options: PathTemplateBasedRequestPrepareOptions | UrlBasedRequestPrepareOptions): Promise<TResult> {
```
bishal-pdMSFT, line 23: why do we need a new sendrequest implementation?

src/configure/clients/github/githubClient.ts
```diff
@@ -36,41 +36,38 @@ export class GithubClient {
...
+                "has_wiki": true
+            },
+            deserializationMapper: null,
+            serializationMapper: null
+        })
+            .then((detail: GitHubRepo) => {
+                return detail;
+            }).catch(error => {
+                telemetryHelper.logError(Layer, TracePoints.GitHubRepositoryCreationFailed, error);
+                if (error.response.statusCode === 422) {
```
bishal-pdMSFT, line 66: when does this happen?

src/configure/clients/github/githubClient.ts
```diff
@@ -36,41 +36,38 @@ export class GithubClient {
...
                     return null;
-                });
-        }
-        catch (error) {
-            telemetryHelper.logError(Layer, TracePoints.GitHubRepositoryCreationFailed, error);
-            return null;
-        }
+                }
```
bishal-pdMSFT, line 68: should you throw an error rather than returning null. This would ensure consistency in the function's behavior in case something goes wrong. Right now it is throowing error in some case and returning null in some.

src/configure/clients/restClient.ts
JA-04: The API call for creating new repo can fail due to multiple reasons, including the existence of repo of the same name. In case of an existing repo name, we need to make API call again with "local repo name + org name" but in case of any other failure, the process needs to be terminated. For deciding the right condition for termination, error code is needed that is why the entire `response` needs to be returned instead of `response.body`.

src/configure/clients/github/githubClient.ts
JA-04: It happens when repo creation fails due to the existence of another repo of same name.

src/configure/clients/github/githubClient.ts
JA-04: `null` is returned only in the case where API call needs to be made again (same repo name exists). If the error is directly thrown then in the calling function there will be no error code to check the reason for failure.

src/configure/clients/github/githubClient.ts
```diff
@@ -36,41 +36,39 @@ export class GithubClient {
...
+                "description": "Repo created from VScode extension 'Deploy to Azure'",
+                "homepage": "https://github.com",
+                "private": true,
+                "has_issues": true,
+                "has_projects": true,
+                "has_wiki": true
+            },
+            deserializationMapper: null,
+            serializationMapper: null,
+            sendCompletetResponse: true
```
bishal-pdMSFT, line 61: rename to `returnFullResponseForFailure`

src/configure/helper/commonHelper.ts
```diff
@@ -93,18 +93,18 @@ export async function executeFunctionWithRetry(
...
-            try {
-                let result = await func();
-                return result;
-            }
-            catch (error) {
-                internalError = error;
-                logger.log(JSON.stringify(error));
-                await Q.delay((resolve) => {resolve();}, retryIntervalTimeInSec * 1000);
-            }
+    let internalError = null;
```
bishal-pdMSFT, line 96: revert this change

src/configure/clients/github/githubClient.ts
JA-04: Done

src/configure/helper/commonHelper.ts
JA-04: Okay

PR #125, opened 2020-07-06
Unit test framework

vineetmimrot:
@JA-04 you can take these changes for running unit tests.


bishal-pdMSFT:
@vineetmimrot should we abandon this?

PR #126, opened 2020-07-07
Add hack to choose first value as default when setting input value ba…

bishal-pdMSFT:
…sed on repo analysis settings


src/configure/templateInputHelper/RepoAnalysisSettingInputProvider.ts
```diff
@@ -69,7 +69,9 @@ export class RepoAnalysisSettingInputProvider {
                     selectedValue = possibleValues[0];
                 }
                 else {
-                    selectedValue = await new ControlProvider().showQuickPick(repoAnalysisSettingKey, possibleValues, { placeHolder: inputControl.getInputDescriptor().name });
+                    // HACK: Temporarily default to first value without asking user. Remove this HACK later and uncomment the next line
+                    selectedValue = possibleValues[0];               
+                    //selectedValue = await new ControlProvider().showQuickPick(repoAnalysisSettingKey, possibleValues, { placeHolder: inputControl.getInputDescriptor().name });
```
kanika1894, line 74: Should we not remove the if condition and anyway set the value of selected value to possibleValues[0]? We can keep a comment to add a else condition later if we plan to change the experience.
Because as of now, both if and else blocks are having the same code.

bishal-pdMSFT: I did that intentionally Kanika. This is a hack and it should be very straight-forward to remove this hack - as simple as uncommenting one line. I felt okay to not have optimized code for this one.

bishal-pdMSFT:
@JA-04 the build is failing, most probably might due to your PR.

```
src/configure/clients/restClient.ts(13,37): error TS2339: Property 'returnFullResponseForFailure' does not exist on type 'PathTemplateBasedRequestPrepareOptions | UrlBasedRequestPrepareOptions'.

```

PR #128, opened 2020-07-09
GitHub repo creation unit tests

JA-04:
Added the following unit tests: 
ImprovedUnitTests


package.json
```diff
@@ -112,27 +112,32 @@
         "compile": "tsc -p ./ && node copyStaticFiles.js",
         "watch": "node copyStaticFiles.js && tsc -watch -p ./",
         "postinstall": "node ./node_modules/vscode/bin/install",
-        "test": "npm run compile && node ./node_modules/vscode/bin/test",
```
bishal-pdMSFT, line 115: This earlier command was pretty interesting as in they go run test binaries from under /node_modules/vscode. What are those test supposed to be?

package.json
```diff
@@ -112,27 +112,32 @@
...
-        "unittest": "node_modules/.bin/mocha -u tdd out/unittest/*.js"
+        "pretest": "npm run compile",
```
bishal-pdMSFT, line 115: what is the use of this command?

package.json
```diff
@@ -112,27 +112,32 @@
...
+        "test": "node ./out/configure/test/runTest.js"
```
bishal-pdMSFT, line 116: can we have one command which builds as well as runs test. May be we can have two commands - `buildAndTest` and `test`

src/configure/test/runTest.ts
```diff
@@ -0,0 +1,23 @@
...
+	try {
+		// The folder containing the Extension Manifest package.json
+		// Passed to `--extensionDevelopmentPath`
+		const extensionDevelopmentPath = path.resolve(__dirname, '../../');
+
+		// The path to test runner
+		// Passed to --extensionTestsPath
+		const extensionTestsPath = path.resolve(__dirname, './suite/index');
+
+		// Download VS Code, unzip it and run the integration test
```
bishal-pdMSFT, line 15: Does it download even if vs code is already installed on machine? Which version does it download?

src/configure/test/runTest.ts
```diff
@@ -0,0 +1,23 @@
+import * as path from 'path';
+
+import { runTests } from 'vscode-test';
+
+async function main() {
```
bishal-pdMSFT, line 5: rename this file to `testHost`

src/configure/test/runTest.ts
```diff
@@ -0,0 +1,23 @@
...
+		const extensionDevelopmentPath = path.resolve(__dirname, '../../');
+
+		// The path to test runner
+		// Passed to --extensionTestsPath
+		const extensionTestsPath = path.resolve(__dirname, './suite/index');
+
+		// Download VS Code, unzip it and run the integration test
+		await runTests({ extensionDevelopmentPath, extensionTestsPath });
+	} catch (err) {
+		console.error('Failed to run tests');
```
bishal-pdMSFT, line 18: can we print how many tests (and their names) failed

src/configure/test/runTest.ts
```diff
@@ -0,0 +1,23 @@
...
+
+async function main() {
+	try {
+		// The folder containing the Extension Manifest package.json
+		// Passed to `--extensionDevelopmentPath`
+		const extensionDevelopmentPath = path.resolve(__dirname, '../../');
+
+		// The path to test runner
+		// Passed to --extensionTestsPath
+		const extensionTestsPath = path.resolve(__dirname, './suite/index');
```
bishal-pdMSFT, line 13: can we pass the test file or the test name which we want to run? This can be useful when we are trying to fix one specific test and don't want to run all tests all the time

src/configure/test/runTest.ts
bishal-pdMSFT: ahh sorry, this is when running the tests itself failed. you should print the error message as well

src/configure/test/suite/index.ts
```diff
@@ -0,0 +1,38 @@
+import * as glob from 'glob';
```
bishal-pdMSFT, line 1: I think this file can be outside `suite` folder as this an index file

src/configure/test/suite/index.ts
```diff
@@ -0,0 +1,38 @@
...
+import * as Mocha from 'mocha';
+import * as path from 'path';
+
+export function run(): Promise<void> {
+	// Create the mocha test
+	const mocha = new Mocha({
+		ui: 'bdd',
```
bishal-pdMSFT, line 8: what does `bdd` mean here?

src/configure/test/suite/index.ts
```diff
@@ -0,0 +1,38 @@
...
+		color: true
+	});
+
+	const testsRoot = path.resolve(__dirname, '..');
+
+	return new Promise((c, e) => {
+		glob('**/**.test.js', { cwd: testsRoot }, (err, files) => {
```
bishal-pdMSFT, line 15: we should only glob for files under `suite` folder

src/configure/test/suite/index.ts
```diff
@@ -0,0 +1,38 @@
...
+				// Run the mocha test
+				mocha.run(failures => {
+					if (failures > 0) {
+						e(new Error(`${failures} tests failed.`));
+					} else {
+						c();
+					}
+				});
+			} catch (err) {
+				console.error(err);
```
bishal-pdMSFT, line 33: Better than logging the error in console here, log it in the caller. that way you don't need to do console.log before everytime you call e()

src/configure/test/suite/GitHubClient.test.ts
```diff
@@ -0,0 +1,135 @@
...
+                orgs.forEach((org) => {
+                    expect(org).to.have.property('login');
+                    expect(org).to.have.property('id');
+                    expect(org).to.have.property('url');
+                });
+                done();
+            });
+        });
+    });
+    before(function () {
```
bishal-pdMSFT, line 54: can we have multiple `before` in one `describe`? I think we should not do that as this could be confusing and could lead to errors. I see three options here:
1. Create a separate `describe` block
2. remove this second `before` and move the `nock` statement to inside the test itself
3. add a function called getOrgList() which will either return valid list or an empty list based on an environment variable. before will use this function inside nock call. each test can set proper environment variables to get desired response.

bishal-pdMSFT: I prefer #3. But this requires that nock support calling a function to get mocked response

src/configure/test/suite/GitHubClient.test.ts
```diff
@@ -0,0 +1,135 @@
...
+var expect = require('chai').expect;
+var nock = require('nock');
+
+let org = "Organization";
+let repoName = "Repository";
+let githubClient = new GithubClient("PAT", "repoUrl");
+
+
+describe('# Testing listOrganizations() ', function () {
+    this.timeout(500000);
```
bishal-pdMSFT, line 12: this is 5s or 500s?

src/configure/test/suite/GitHubClient.test.ts
```diff
@@ -0,0 +1,135 @@
...
+        });
+    });
+    before(function () {
+        var response = [];
+        nock('https://api.github.com')
+            .get('/user/orgs')
+            .reply(200, response);
+    });
+
+    context('User is not a member of any organization', function () {
```
bishal-pdMSFT, line 61: please add some negative cases - when user does not have permission

src/configure/test/suite/GitHubClient.test.ts
```diff
@@ -0,0 +1,135 @@
...
+
+        nock('https://api.github.com')
+            .post('/orgs/' + org + '/repos')
+            .reply(200, response);
+    });
+
+    context('Given repository name is unique', function () {
+        it('Should create a new repository', function (done) {
+            let githubClient = new GithubClient("PAT", "repoUrl");
+            githubClient.createGithubRepo(org, repoName).then((repo) => {
```
bishal-pdMSFT, line 96: this test will always succeed even if product code does not send org and repo name to the API correctly as you can mocking those things in the response.
While creating mocked response, you should make sure that some properties are filled up based on the api payload.

src/configure/test/suite/GitHubClient.test.ts
```diff
@@ -0,0 +1,135 @@
+import { GithubClient } from "../../clients/github/githubClient";
```
bishal-pdMSFT, line 1: this class is a very low level class and while it is good to have tests for these, it would be better to also have tests for the classes where these client functions are called?

src/configure/test/suite/GitHubClient.test.ts
```diff
@@ -0,0 +1,135 @@
...
+                expect(repo).to.have.property('html_url');
+                expect(repo).to.have.property('owner');
+                expect(repo.name).to.deep.equal(repoName);
+                expect(repo.description).to.deep.equal("Repo created from VScode extension 'Deploy to Azure'");
+                done();
+            });
+        });
+    });
+
+    before(function () {
```
bishal-pdMSFT, line 108: as noted, please re-think about such `before`s

src/configure/test/suite/index.ts
JA-04: Okay, will change it.

src/configure/test/suite/GitHubClient.test.ts
JA-04: Option 1 - creating a separate describe block for each test seems like the best option because the `nock` interceptor needs to be defined inside `before` block due to which 2nd option is not feasible, and 3rd option might not provide the clarity of code which we are trying to target.

src/configure/test/suite/GitHubClient.test.ts
JA-04: Added.

src/configure/test/suite/GitHubClient.test.ts
JA-04: Okay

src/configure/test/suite/GitHubClient.test.ts
JA-04: Actually I have written tests for both, functions of this class and as well as for the functions which are calling the functions of this class.

src/configure/test/suite/GitHubClient.test.ts
JA-04: But the main purpose behind putting multiple tests and hence multiple `before`s inside single `describe` was to create a set of tests that test different cases of a single function. Should I change this approach?

package.json
JA-04: This command is used for building the project.

package.json
JA-04: When we push the code, checks are already there for checking build related issues. Therefore, I created a separate command which only runs the tests.

src/configure/test/runTest.ts
JA-04: No, it only downloads if it not already downloaded. The version which it downloads it is 1.47.1

src/configure/test/runTest.ts
JA-04: Okay

src/configure/test/runTest.ts
JA-04: Okay, will add it.

src/configure/test/suite/GitHubClient.test.ts
JA-04: 500s, but this was put because earlier tests were taking longer to execute but now that issue is resolved so I'll change it back to default timeout value.

src/configure/test/suite/index.ts
JA-04: It is the name of the interface. BDD = Behaviour Driven Development.

src/configure/test/suite/GitHubClient.test.ts
JA-04: Noted, will make the required changes.

src/configure/test/runTest.ts
JA-04: For running a particular test, we can use the command `mocha -g 'test_name'`.

src/configure/test/runTest.ts
```diff
@@ -0,0 +1,23 @@
...
+		// The folder containing the Extension Manifest package.json
+		// Passed to `--extensionDevelopmentPath`
+		const extensionDevelopmentPath = path.resolve(__dirname, '../../');
+
+		// The path to test runner
+		// Passed to --extensionTestsPath
+		const extensionTestsPath = path.resolve(__dirname, './index');
+
+		// Download VS Code, unzip it and run the integration test
+		await runTests({ extensionDevelopmentPath, extensionTestsPath });
```
bishal-pdMSFT, line 16: try passing optional test file name as extensionTestsEnv

bishal-pdMSFT:
Please update the README with test information:
1. How to run tests
2. How to add a test

PR #134, opened 2020-07-30
Adding notification post GitHubPat input

kanika1894:
Since the user isn't aware of the repo analysis call being made post the input of GitHubPat, it is likely that he will try creating a new flow using Ctrl+Shift+P which creates faulty telemetry logs. Therefore, adding additional notification to prevent such scenarios.

### DIFF
--- a/src/configure/resources/messages.ts
+++ b/src/configure/resources/messages.ts
@@ -108,4 +108,5 @@ export class Messages {
     public static cannotCreateGitHubRepository = "New GitHub repository could not be created as the repository with the same name already exists";
     public static languageNotSupported = "The language of the repository selected is not supported.";
     public static UnableToGetTemplateParameters = "Unable to get parameters for the selected template.";
+    public static FetchingRepoDetails = "Fetching repository details";
 }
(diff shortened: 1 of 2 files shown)


bishal-pdMSFT:
@kanika1894 we already had notification for repo analysis. Has that regressed?

kanika1894:
> @kanika1894 we already had notification for repo analysis. Has that regressed?

No, that is still there
In configure class, there is a ``` getSelectedPipeline() ``` function which has that notification - Analyzing your repo
But that comes quite late in the flow, and till there is nothing for user display.
The time difference is more specially when the gitHubPat entered is incorrect.

bishal-pdMSFT:
okay got it. so that notification is for local analysis.

I think we should change that notification to say "fetching templates"

src/configure/resources/messages.ts
```diff
@@ -108,4 +108,5 @@ export class Messages {
     public static cannotCreateGitHubRepository = "New GitHub repository could not be created as the repository with the same name already exists";
     public static languageNotSupported = "The language of the repository selected is not supported.";
     public static UnableToGetTemplateParameters = "Unable to get parameters for the selected template.";
+    public static FetchingRepoDetails = "Fetching repository details";
```
bishal-pdMSFT, line 111: String could be `Analyzing your repository`

kanika1894: updated

PR #140, opened 2020-08-10
Not supported repo providers 

kanika1894:
Since we support tfs and gitHub repo providers, this will let us improve insights on the failures due to this issue.
Though the status of pipeline will be failed (since the flow is breaking in between), the whitlelisted key will help us have better analysis.

### DIFF
--- a/src/configure/configure.ts
+++ b/src/configure/configure.ts
@@ -64,11 +64,11 @@ export async function configurePipeline(node: AzureTreeItem) {
         }
         catch (error) {
             if (!(error instanceof UserCancelledError)) {
+                vscode.window.showErrorMessage(error.message);
                 if (extensionVariables.isErrorWhitelisted === true) {
                     telemetryHelper.setResult(Result.Succeeded);
                 } else {
                     extensionVariables.outputChannel.appendLine(error.message);
-                    vscode.window.showErrorMessage(error.message);
                     telemetryHelper.setResult(Result.Failed, error);
                 }
             }
--- a/src/configure/configure.ts
+++ b/src/configure/configure.ts
@@ -405,7 +405,7 @@ class Orchestrator {
                 };
             }
             else {
-                let repositoryProvider: string = "Other";
+                let repositoryProvider: string;
 
                 if (remoteUrl.indexOf("bitbucket.org") >= 0) {
                     repositoryProvider = "Bitbucket";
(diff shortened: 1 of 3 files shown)


src/configure/configure.ts
```diff
@@ -405,16 +405,20 @@ class Orchestrator {
...
                 }
                 else if (remoteUrl.indexOf("gitlab.com") >= 0) {
                     repositoryProvider = "GitLab";
                 }
+                else {
+                    repositoryProvider = remoteUrl;
+                }
 
                 telemetryHelper.setTelemetry(TelemetryKeys.RepoProvider, repositoryProvider);
+                telemetryHelper.setTelemetry(TelemetryKeys.IsErrorWhitelisted, "true");
```
vineetmimrot, line 421: You should also put a if condition for IsErrorWhitelisted on line 66 and don't set the result to failed.

kanika1894: So basically have a variable IsErrorWhitelisted and put condition according to that?
(I though our query for results will have the condition based on the telemetry key)

vineetmimrot: Yes have a variable and put failed/succeed condition acc to that and also set the whitelisted telemetry values. So by default we will get all the whitelisted error on succeeded and if we need to dig down further in future we can use the isWhitelisted flag in kusto.

kanika1894: updated

src/configure/configure.ts
```diff
@@ -64,9 +64,13 @@ export async function configurePipeline(node: AzureTreeItem) {
         }
         catch (error) {
             if (!(error instanceof UserCancelledError)) {
-                extensionVariables.outputChannel.appendLine(error.message);
-                vscode.window.showErrorMessage(error.message);
-                telemetryHelper.setResult(Result.Failed, error);
+                if (extensionVariables.isErrorWhitelisted === true) {
```
vineetmimrot, line 67: We should show error to the user in this case also.

src/configure/configure.ts
```diff
@@ -64,9 +64,13 @@ export async function configurePipeline(node: AzureTreeItem) {
         }
         catch (error) {
             if (!(error instanceof UserCancelledError)) {
-                extensionVariables.outputChannel.appendLine(error.message);
                 vscode.window.showErrorMessage(error.message);
-                telemetryHelper.setResult(Result.Failed, error);
+                if (extensionVariables.isErrorWhitelisted === true) {
+                    telemetryHelper.setResult(Result.Succeeded);
+                } else {
+                    extensionVariables.outputChannel.appendLine(error.message);
```
vineetmimrot, line 71: what is this for? any idea?

kanika1894: Not sure about the role, but since found at every place where result is being set, have made it common here as well.

Issue #137, opened 2020-08-06
Cannot read property 'session' of undefined (v1.1.1)


issacnitin:
![Capture](https://user-images.githubusercontent.com/4218887/89547270-2c7bfe80-d823-11ea-9b8c-06744fc4ec05.PNG)

Casually checking in 😃 
Happens after logging in to Azure

vineetmimrot:
Thanks for reporting man, @Kanika<mailto:Kanika.Pasrija@microsoft.com> can you please look into it. 😊
Can you please tell which version of Azure account extension you are using?

--
Best Regards,
Vineet Mimrot

bishal-pdMSFT:
Thanks @issacnitin :)
@kanika1894 please check thia

issacnitin:
> Thanks for reporting man, @kanika<mailto:Kanika.Pasrija@microsoft.com> can you please look into it. 😊 Can you please tell which version of Azure account extension you are using?
> […](#)
[...]

PR #141, opened 2020-08-10
Fix session undefined bug

vineetmimrot:
After the Azure Login, from the investigation I found that it takes time to update the subscription list. So I have added timeout for the same.

The bug is reproducible by:
1. Do Azure signout
2. Remove Deploy to Azure and Azure Account extension.
3. Install Deploy to Azure extension.
4. Try to setup pipeline, for the first time you will hit "session" is undefined. Though from second time onwards things work fine.

### DIFF
--- a/src/configure/configure.ts
+++ b/src/configure/configure.ts
@@ -171,7 +171,7 @@ class Orchestrator {
         if (this.continueOrchestration) {
             await this.getSourceRepositoryDetails();
             if (!this.inputs.azureSession) {
-                this.inputs.azureSession = await getAzureSession();
+                this.inputs.azureSession = getAzureSession();
             }
             let repoAnalysisResult = await this.getRepositoryAnalysis();
             await this.getSelectedPipeline(repoAnalysisResult);
--- a/src/configure/configure.ts
+++ b/src/configure/configure.ts
@@ -40,12 +40,13 @@ export async function configurePipeline(node: AzureTreeItem) {
                 // set telemetry
                 telemetryHelper.setTelemetry(TelemetryKeys.AzureLoginRequired, 'true');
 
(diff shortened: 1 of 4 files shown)


vineetmimrot:
https://github.com/microsoft/vscode-deploy-azure/issues/137

vineetmimrot:
Raised issue
https://github.com/microsoft/vscode-azure-account/issues/210

src/configure/configure.ts
```diff
@@ -40,12 +40,13 @@ export async function configurePipeline(node: AzureTreeItem) {
...
                 telemetryHelper.setTelemetry(TelemetryKeys.AzureLoginRequired, 'true');
 
-                let loginOption = await vscode.window.showInformationMessage(Messages.azureLoginRequired, Messages.signInLabel, Messages.signUpLabel);
+                const loginOption = await vscode.window.showInformationMessage(Messages.azureLoginRequired, Messages.signInLabel, Messages.signUpLabel);
                 if (loginOption && loginOption.toLowerCase() === Messages.signInLabel.toLowerCase()) {
                     telemetryHelper.setTelemetry(TelemetryKeys.AzureLoginOption, 'SignIn');
                     await vscode.window.withProgress({ location: vscode.ProgressLocation.Notification, title: Messages.waitForAzureSignIn },
                         async () => {
                             await vscode.commands.executeCommand("azure-account.login");
+                            await extensionVariables.azureAccountExtensionApi.waitForSubscriptions();
```
kanika1894, line 49: So this is basically ensuring that we have the list of subscriptions with us and don't run into the error, right?

vineetmimrot: yes.

PR #144, opened 2020-08-22
Users/nikavima/vscode

NinadKavimandan:
### Adding a new Asset - DotNetCoreSdkVersionGetter
This returns the .NET Core version as "DOTNETCORE|{version}", compliant with supported linuxFxVersion values.

Also removing the hack to take first value as default for working directory


src/configure/templateInputHelper/RepoAnalysisSettingInputProvider.ts
```diff
@@ -69,9 +69,7 @@ export class RepoAnalysisSettingInputProvider {
                     selectedValue = possibleValues[0];
                 }
                 else {
-                    // HACK: Temporarily default to first value without asking user. Remove this HACK later and uncomment the next line
-                    selectedValue = possibleValues[0];               
-                    //selectedValue = await new ControlProvider().showQuickPick(repoAnalysisSettingKey, possibleValues, { placeHolder: inputControl.getInputDescriptor().name });
+                    selectedValue = await new ControlProvider().showQuickPick(repoAnalysisSettingKey, possibleValues, { placeHolder: inputControl.getInputDescriptor().name });
```
kanika1894, line 72: Removing the hack means we'll now be having a different user experience then?

bishal-pdMSFT: yep this is tricky. 
Kanika does this hack currently hide even the question about choosing working directory? I think the intention was hide the question only for multiple secondary files (like webpack.config files). We should still be asking user to choose a working directory if multiple exist.

kanika1894: yeah, we are asking the user for the working directory if multiple exist
This was hiding secondary files, node version, basically the parameters which could have multiple values, to keep the user experience same.

src/configure/utilities/webAppDotNetCoreVersionConverter.ts
```diff
@@ -0,0 +1,3 @@
+export async function webAppDotNetCoreRuntimeConverter(dotnetcoreVersion: string) {
+    return "DOTNETCORE|" + dotnetcoreVersion.replace("netcoreapp", "");
```
bishal-pdMSFT, line 2: this need not be an asset. you can use a string manipulation mustache helper to achieve this.

src/configure/templateInputHelper/RepoAnalysisSettingInputProvider.ts
bishal-pdMSFT: Thanks @kanika1894. @NinadKavimandan this is in contrast to what you mentioned. you said you were not even asked the working directory. can you please re-check?

NinadKavimandan: @bishal-pdMSFT @kanika1894 it's still picking up first folder from `ApplicationSettings` for me when i ran the extension locally. For the shipped extension, its somehow failing at `Select organization` step. :(

kanika1894: @bishal-pdMSFT I checked again
Since we made working directory as a parameter (coming from template) while integrating with template service, the support of choosing working directory got lost for web app scenarios. Since AKS support is still the one local to VSCode, the option to select working directory still exists for AKS scenario.

And I guess when we introduced the hack, that time also we had a discussion that the option to choose working directory is something which we should have but since as per Pulkit, we wanted no change for user experience at that stage, we kept that hack considering we'll remove this sometime later.

src/configure/utilities/webAppDotNetCoreVersionConverter.ts
NinadKavimandan: using mustache helper instead

src/configure/templateInputHelper/RepoAnalysisSettingInputProvider.ts
bishal-pdMSFT: @kanika1894 ah yes working directory as a template parameter means that the input moves to after azure resource is selected. that was a problem and the hack should fix it. is that right understanding?

@NinadKavimandan If so, we should not revert the hack without Pulkit's approval. Kanika can you please ask Pulkit if he is fine to remove this hack now.

NinadKavimandan: @bishal-pdMSFT @kanika1894 reverted the removal of hack with latest push

bishal-pdMSFT (approved):

bishal-pdMSFT:
@NinadKavimandan is this PR good to go?

NinadKavimandan:
@bishal-pdMSFT yes, was holding on in case any change needs to be pushed for .NET framework.

bishal-pdMSFT (approved):

PR #145, opened 2020-08-25
Show personal account in org list

vineetmimrot:
This PR appends GitHub owner account into org list. Many users doesn't have org, so this feature will allow users to setup workflow in their default account.


bishal-pdMSFT:
@vineetmimrot a better description please. We are in general trying to do that across PRs :)

vineetmimrot:
updated description.

bishal-pdMSFT (approved):

PR #146, opened 2020-09-02
fix Azure pipeline with GitHub repo fails with error bug

vineetmimrot:
[Task 1757569](https://dev.azure.com/mseng/AzureDevOps/_boards/board/t/Deployment-Azure/Stories/?workitem=1757569): Bug - Azure pipeline with GitHub repo fails with error

[Task 1766390](https://dev.azure.com/mseng/AzureDevOps/_boards/board/t/Deployment-Azure/Stories/?workitem=1766390): Bug: User is not authorized to access this resource

### DIFF
--- a/src/configure/configure.ts
+++ b/src/configure/configure.ts
@@ -1,6 +1,5 @@
 import { GenericResource } from 'azure-arm-resource/lib/resource/models';
 import { ApplicationSettings, RepositoryAnalysis } from 'azureintegration-repoanalysis-client-internal';
-import * as crypto from 'crypto';
 import * as path from 'path';
 import * as vscode from 'vscode';
 import { AzureTreeItem, UserCancelledError } from 'vscode-azureextensionui';
@@ -28,6 +27,7 @@ import { Messages } from './resources/messages';
 import { TelemetryKeys } from './resources/telemetryKeys';
 import { TracePoints } from './resources/tracePoints';
 import { InputControlProvider } from './templateInputHelper/InputControlProvider';
+import { Utilities } from './utilities/utilities';
 
 const uuid = require('uuid/v4');
 
@@ -140,8 +140,7 @@ class Orchestrator {
             this.inputs.azureSession.environment, this.inputs.azureSession.tenantId, this.inputs.subscriptionId);
         telemetryHelper.setTelemetry(TelemetryKeys.resourceType, this.inputs.targetResource.resource.type);
         telemetryHelper.setTelemetry(TelemetryKeys.resourceKind, this.inputs.targetResource.resource.kind);
-        telemetryHelper.setTelemetry(TelemetryKeys.resourceIdHash,
-            crypto.createHash('sha256').update(this.inputs.targetResource.resource.id).digest('hex'));
+        telemetryHelper.setTelemetry(TelemetryKeys.resourceIdHash, Utilities.createSha256Hash(this.inputs.targetResource.resource.id));
     }
 
     private async selectTemplate(resource: GenericResource): Promise<void> {
--- a/src/configure/configure.ts
+++ b/src/configure/configure.ts
@@ -27,7 +27,7 @@ import { Messages } from './resources/messages';
 import { TelemetryKeys } from './resources/telemetryKeys';
 import { TracePoints } from './resources/tracePoints';
 import { InputControlProvider } from './templateInputHelper/InputControlProvider';
-import { Utilities } from './utilities/utilities';
+import { Utilities, WhiteListedError } from './utilities/utilities';
 
 const uuid = require('uuid/v4');
 
@@ -68,7 +68,7 @@ export async function configurePipeline(node: AzureTreeItem) {
             if (!(error instanceof UserCancelledError)) {
                 vscode.window.showErrorMessage(error.message);
                 extensionVariables.outputChannel.appendLine(error.message);
-                if (extensionVariables.isErrorWhitelisted === true) {
+                if (error instanceof WhiteListedError) {
                     telemetryHelper.setTelemetry(TelemetryKeys.IsErrorWhitelisted, "true");
                     telemetryHelper.setResult(Result.Succeeded, error);
                 } else {
@@ -425,9 +425,8 @@ class Orchestrator {
                     repositoryProvider = remoteUrl;
                 }
 
-                extensionVariables.isErrorWhitelisted = true;
                 telemetryHelper.setTelemetry(TelemetryKeys.RepoProvider, repositoryProvider);
-                throw new Error(Messages.cannotIdentifyRespositoryDetails);
+                throw new WhiteListedError(Messages.cannotIdentifyRespositoryDetails);
             }
         }
         else {
--- a/src/configure/configure.ts
+++ b/src/configure/configure.ts
@@ -1,5 +1,6 @@
 import { GenericResource } from 'azure-arm-resource/lib/resource/models';
 import { ApplicationSettings, RepositoryAnalysis } from 'azureintegration-repoanalysis-client-internal';
+import * as crypto from 'crypto';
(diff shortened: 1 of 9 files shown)


bishal-pdMSFT:
@kanika1894 please review this

src/configure/configurers/azurePipelineConfigurer.ts
```diff
@@ -103,6 +103,10 @@ export class AzurePipelineConfigurer implements Configurer {
             telemetryHelper.setTelemetry(TelemetryKeys.NewOrganization, inputs.isNewOrganization.toString());
         }
         catch (error) {
+            if ((error.typeKey as string).toUpperCase() === constants.ExceptionType.UnauthorizedRequestException) {
+                extensionVariables.isErrorWhitelisted = true;
```
bishal-pdMSFT, line 107: this is tricky Vineet. you are whitelisting an error by setting a global variable. What if this error is suppressed by the caller (down the stack) and another error is thrown which is not expected. Now what will happen is that the 2nd error will get whitelisted. 
I think the intention here is to say that this error is resulting in a failure which should be whitelisted. There are two ways which comes to my mind currently:
1. Rename to `isFailureWhitelisted` and ensure that execution ends with this error. Essentially we are saying that this error is resulting into a failure. I am not sure how we will enforce this.
2. Extend error object to include a variable `isExpected`. This changes the scope from global to local. Now whenever some throws an error they can decide if this is an expected error. At top level, we will check this property of error object and whitelist accodingly.

I do not have a clear preference. But I think 2nd one is bit better as it has lesser chance of whitelisting being missed/misused.

src/configure/resources/messages.ts
```diff
@@ -110,4 +110,5 @@ export class Messages {
     public static UnableToGetTemplateParameters = "Unable to get parameters for the selected template.";
     public static AnalyzingRepo = "Analyzing your repository";
     public static AzureLoginError = "Error in getting Azure login information, Please try again.";
+    public static AdoDifferentTenantError = " The AAD tenant of your Azure DevOps repository is different than the one you have been logged-in to VSCode. Open 'Command Pallete' and select 'Azure: Sign Out' and then Sign-In to other tenant by invoking 'Deploy to Azure' extension";
```
bishal-pdMSFT, line 113: How sure are we that unauthorized is due to different tenant? if not 100%, then we should word the message a bit vaguely. e.g. One reason could be that the AAD tenant of your Azure DevOps repostiory is different than ...

src/configure/configure.ts
```diff
@@ -110,7 +112,7 @@ class Orchestrator {
             await pipelineConfigurer.createPreRequisites(this.inputs, !!this.azureResourceClient ? this.azureResourceClient : new AppServiceClient(this.inputs.azureSession.credentials, this.inputs.azureSession.environment, this.inputs.azureSession.tenantId, this.inputs.subscriptionId));
 
             telemetryHelper.setCurrentStep('CreateAssets');
-            if (this.inputs.pipelineConfiguration.template.templateType === TemplateType.REMOTE) {
+            if (this.inputs.pipelineConfiguration.template.templateType === TemplateType.REMOTE && this.inputs.sourceRepository.repositoryProvider === RepositoryProvider.Github) {
```
bishal-pdMSFT, line 115: please explain the fix. The bug says that this happens when repo provider is github. But with the change, the flow will still go in this block.

src/configure/configure.ts
```diff
@@ -138,6 +140,8 @@ class Orchestrator {
             this.inputs.azureSession.environment, this.inputs.azureSession.tenantId, this.inputs.subscriptionId);
         telemetryHelper.setTelemetry(TelemetryKeys.resourceType, this.inputs.targetResource.resource.type);
         telemetryHelper.setTelemetry(TelemetryKeys.resourceKind, this.inputs.targetResource.resource.kind);
+        telemetryHelper.setTelemetry(TelemetryKeys.resourceIdHash,
+            crypto.createHash('sha256').update(this.inputs.targetResource.resource.id).digest('hex'));
```
bishal-pdMSFT, line 144: move the crypto to a utility.

src/configure/resources/constants.ts
```diff
@@ -196,4 +196,8 @@ export const ingressManifest: string = "ingress";
 export const azurePipeline: string = "Azure-pipeline";
 export const githubWorkflow: string = "Github-workflow";
 export const clientPropertyKey: string = "ms.client.vscode";
-export const inputModeProperty: string = "inputMode";
\ No newline at end of file
+export const inputModeProperty: string = "inputMode";
+
+export const ExceptionType = {
+    UnauthorizedRequestException: 'UNAUTHORIZEDREQUESTEXCEPTION'
```
bishal-pdMSFT, line 202: do you want to use underscores to make it more readable

src/configure/configurers/azurePipelineConfigurer.ts
vineetmimrot: Yeah, I also thought earlier about 2nd approach but didn't found any scenario in our code where the error is suppressed by the caller. Yeah, I will implement this approach.

src/configure/resources/messages.ts
vineetmimrot: This error is happening while calling "_apis/git/repositories/{repo_name}" API and the user has already cloned the repo locally. So even if he has READER privilege, he must be able to do get over this API. The only other reason I can think of is that when the user has been removed from the project. 

Yeah, I will update the text.

src/configure/configure.ts
vineetmimrot: The bug says -> Repo provider is ADO and repository is GitHub then this issue is happening.

src/configure/resources/constants.ts
vineetmimrot: The right side value is the exception came from the API response. To avoid case ignore comparison, I have converted into upper case.

src/configure/configure.ts
vineetmimrot: done

src/configure/configure.ts
kanika1894: What if the repoProvider is GitHub but ``` extensionVariables.enableGitHubWorkflow ``` is false
Is this fix handling thise scenario? If not, then the condition can be on the type of pipelineConfigurer insread of repo provider.

src/configure/configurers/azurePipelineConfigurer.ts
vineetmimrot: done

src/configure/configure.ts
vineetmimrot: @kanika1894 
if extensionVariables.enableGitHubWorkflow is false, then how can repositoryProvider is GitHub?
I think you are confusing repositoryProvider with repo type.

src/configure/utilities/utilities.ts
```diff
@@ -0,0 +1,21 @@
+import * as crypto from 'crypto';
+
+export class Utilities {
+    public static createSha256Hash(input: string): string {
+        return crypto.createHash('sha256').update(input).digest('hex');
+    }
+}
+
+export class WhiteListedError extends Error {
+    constructor(message?: string, error?: any) {
```
bishal-pdMSFT, line 10: do we really need to clone the error? Will this work?
```
export class WhiteListedError extends Error {
    constructor(message?: string) {
        super(message);
    }
}
```
And we can rely on type casting to convert from `Error` to `WhiteListedError`

bishal-pdMSFT (approved):

src/configure/utilities/utilities.ts
vineetmimrot: I want to preserve the actual stacktrace, thus I am cloning the error.

PR #148, opened 2020-09-04
Moda and portal extension clients for provisioning service

tiwarishub:


src/configure/clients/ProvisioningServiceClient.ts
```diff
@@ -0,0 +1,38 @@
...
+import { IProvisioningServiceClient } from "./IProvisioningServiceClient";
+import { RestClient } from "./restClient";
+
+
+export class ProvisioningServiceClient implements IProvisioningServiceClient {
+  private restClient: RestClient;
+  private url: string;
+  constructor( url: string, credentials: ServiceClientCredentials) {
+    this.restClient = new RestClient(credentials);
+    this.url = url;
```
anuragc617, line 12: rename to baseUrl

src/configure/clients/IProvisioningServiceClient.ts
```diff
@@ -0,0 +1,6 @@
+import { ProvisioningConfiguration } from "../model/provisioningConfiguration";
+
+export interface IProvisioningServiceClient {
+    createProvisioningConfiguration(provisioningConfiguration: ProvisioningConfiguration, relativeUrl: string): Promise<ProvisioningConfiguration>;
```
anuragc617, line 4: pass reopId instead of relative url

src/configure/clients/ProvisioningServiceClient.ts
tiwarishub: done

src/configure/clients/IProvisioningServiceClient.ts
tiwarishub: done

src/configure/clients/ProvisioningServiceClient.ts
```diff
@@ -0,0 +1,46 @@
...
+  private restClient: RestClient;
+  private baseUrl: string;
+  private defaultHeaders: { [propertyName: string]: string };
+  private readonly pipelineProvisioningJob = "PipelineProvisioningJob";
+  constructor( baseUrl: string, githubPAT: string, credentials: ServiceClientCredentials) {
+    this.restClient = new RestClient(credentials);
+    this.baseUrl = baseUrl;
+    this.defaultHeaders = {
+      "Content-Type": "application/json",
+      "X-GITHUB-TOKEN": githubPAT,
```
anuragc617, line 16: pat token header is only required if baseUrl is PE service otherwise it will go as Authorization header.
Also api-version query version is needed for PE service.

tiwarishub: okay, then i will add Headers as part of constructor . For api-version , we have optional param queryParam createProvisioning/getProvisioning request which can be used to pass api-version

anuragc617: api-version should be part of ProvisioningServiceClient.ts, consumer of this should not worry about the api version. You can store it in default query params based on baseUrl.

tiwarishub: Updated with changes

src/configure/clients/ProvisioningServiceClient.ts
```diff
@@ -0,0 +1,58 @@
...
+  constructor( serviceDefinition: IServiceUrlDefinition, headers: { [propertyName: string]: string }, credentials: ServiceClientCredentials) {
+    this.restClient = new RestClient(credentials);
+    this.serviceDefinition = serviceDefinition;
+    this.defaultHeaders = headers;
+  }
+
+  public async createProvisioningConfiguration(provisioningConfiguration: ProvisioningConfiguration, githubOrg: string, repositoryId: string): Promise<ProvisioningConfiguration> {
+    const requestUrl = this.serviceDefinition.serviceUrl + githubOrg + "/" + repositoryId + "/" + this.pipelineProvisioningJob;
+    // tslint:disable-next-line:prefer-const
+    let queryParams: { [propertyName: string]: string };
```
anuragc617, line 23: you can create default query params in constructor and use it here and in other function

anuragc617 (approved):

PR #149, opened 2020-09-07
Adding support for array repo analysis setting

anuragc617:
This is the change to show user dropdown, if the repo analysis setting is an array and parameter type is string

### DIFF
--- a/src/configure/templateInputHelper/RepoAnalysisSettingInputProvider.ts
+++ b/src/configure/templateInputHelper/RepoAnalysisSettingInputProvider.ts
@@ -1,6 +1,7 @@
 import { ApplicationSettings } from "azureintegration-repoanalysis-client-internal";
 import { ControlProvider } from "../helper/controlProvider";
 import { telemetryHelper } from "../helper/telemetryHelper";
+import { InputDataType } from "../model/Contracts";
 import { TracePoints } from "../resources/tracePoints";
 import { InputControl } from "./InputControl";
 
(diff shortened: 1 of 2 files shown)


src/configure/templateInputHelper/RepoAnalysisSettingInputProvider.ts
```diff
@@ -71,10 +88,16 @@ export class RepoAnalysisSettingInputProvider {
                 else {
                     selectedValue = await new ControlProvider().showQuickPick(repoAnalysisSettingKey, possibleValues, { placeHolder: inputControl.getInputDescriptor().name });
                 }
-
                 if (settingIndexMap.get(selectedValue.data).length === 1) {
                     this._selectedRepoAnalysisSettingIndex = settingIndexMap.get(selectedValue.data)[0];
                 }
+
+                let repoAnalysisValue = this._repoAnalysisSettings[settingIndexMap.get(selectedValue.data)[0]].settings[repoAnalysisSettingKey];
```
bishal-pdMSFT, line 95: shouldn't this be done earlier when we are showing the drop down at L89? Why do you want to show the drop down twice?

src/configure/templateInputHelper/RepoAnalysisSettingInputProvider.ts
```diff
@@ -71,10 +88,16 @@ export class RepoAnalysisSettingInputProvider {
                 else {
                     selectedValue = await new ControlProvider().showQuickPick(repoAnalysisSettingKey, possibleValues, { placeHolder: inputControl.getInputDescriptor().name });
                 }
-
                 if (settingIndexMap.get(selectedValue.data).length === 1) {
                     this._selectedRepoAnalysisSettingIndex = settingIndexMap.get(selectedValue.data)[0];
```
bishal-pdMSFT, line 92: this also need to be set in case of array

src/configure/templateInputHelper/RepoAnalysisSettingInputProvider.ts
anuragc617: @bishal-pdMSFT this is the case when there multiple repository analysis setting and the setting key value is an array and the value for that setting key to be used is string. So here first dropdown allows the value to be picked from various repository analysis settings and second dropdown is to get the value for the setting.

src/configure/templateInputHelper/RepoAnalysisSettingInputProvider.ts
anuragc617: this is for setting up repository analysis setting in use.

bishal-pdMSFT (approved):

PR #151, opened 2020-09-11
Fixing build

NinadKavimandan:


bishal-pdMSFT:
why so many commits are showing up?

NinadKavimandan:
> why so many commits are showing up?

I added a new remote, as this repo, to the existing branch which had origin as forked repo earlier. So this entire branch was pushed from start because of that. Although the difference from master is not these many commits, can it be the case that after changing the remote it changed commit ids?

bishal-pdMSFT:
> > why so many commits are showing up?
> 
> I added a new remote, as this repo, to the existing branch which had origin as forked repo earlier. So this entire branch was pushed from start because of that. Although the difference from master is not these many commits, can it be the case that after changing the remote it changed commit ids?

yeah, better start with a new clone

bishal-pdMSFT (approved):

NinadKavimandan:
Yes. Thanks!

PR #154, opened 2020-09-16
Provisioning service integration

tiwarishub:


src/configure/configurers/remoteConfigurer.ts
```diff
@@ -0,0 +1,190 @@
...
+
+// tslint:disable-next-line:interface-name
+interface DraftFile {
+    content: string;
+    path: string;
+    absPath: string;
+}
+
+const Layer: string = "RemoteConfigurer";
+export class RemoteConfigurer implements IRemoteConfigurerBase {
```
vineetmimrot, line 26: Should we name it to Provisioning Configurer?

src/configure/configurers/remoteConfigurer.ts
```diff
@@ -0,0 +1,190 @@
...
+
+    public async postSteps(provisioningConfiguration: ProvisioningConfiguration, draftPipelineConfiguration: DraftPipelineConfiguration, inputs: WizardInputs): Promise<void> {
+        await this.getFileToCommit(draftPipelineConfiguration);
+        await this.showPipelineFiles();
+        let displayMessage = Messages.modifyAndCommitFile;
+        if (this.filesToCommit.length > 1) {
+            displayMessage = Messages.modifyAndCommitMultipleFiles;
+        }
+
+        const commitOrDiscard = await vscode.window.showInformationMessage(
```
vineetmimrot, line 92: Use control provider showInformationMessage

src/configure/configurers/remoteConfigurer.ts
```diff
@@ -0,0 +1,190 @@
...
+                    telemetryHelper.setTelemetry(TelemetryKeys.BrowsePipelineClicked, 'true');
+                    vscode.env.openExternal(vscode.Uri.parse(this.queuedPipelineUrl));
+                }
+            });
+    }
+
+    public async postSteps(provisioningConfiguration: ProvisioningConfiguration, draftPipelineConfiguration: DraftPipelineConfiguration, inputs: WizardInputs): Promise<void> {
+        await this.getFileToCommit(draftPipelineConfiguration);
+        await this.showPipelineFiles();
+        let displayMessage = Messages.modifyAndCommitFile;
```
tiwarishub, line 87: ternary operator

src/configure/configurers/remoteConfigurer.ts
```diff
@@ -0,0 +1,190 @@
...
+                throw error;
+            }
+        });
+        telemetryHelper.setTelemetry(TelemetryKeys.WorkflowFileName, fileName);
+        return Path.join(directoryPath, fileName);
+    }
+
+    public async CreatePreRequisiteParams(wizardInputs: WizardInputs): Promise<void>{
+        const parsedCredentials: { [key: string]: any } = JSON.parse(JSON.stringify(wizardInputs.azureSession.credentials));
+        wizardInputs.pipelineConfiguration.params["armAuthToken"] = "Bearer " + parsedCredentials["tokenCache"]["target"]["_entries"][0]["accessToken"];
```
vineetmimrot, line 170: Can you try parsing via https://github.com/Azure/azure-sdk-for-node/blob/master/runtime/ms-rest/lib/credentials/tokenCredentials.js

Actually credentials are of type https://github.com/Azure/azure-sdk-for-node/blob/master/runtime/ms-rest/lib/serviceClient.js#L124

tiwarishub: Yes, we tried with TokenCredentials but in its d.ts file, we have no way to access token param from this class. So last option left was parse it to [string]any then take value from it. We can have a chat/call to discuss this.

src/configure/configurers/remoteConfigurer.ts
tiwarishub: done

src/configure/configure.ts
```diff
@@ -137,19 +143,20 @@ class Orchestrator {
...
         this.inputs.targetResource.resource = await azureResourceSelector.getAzureResource(this.inputs);
         this.azureResourceClient = ResourceSelectorFactory.getAzureResourceClient(targetType, this.inputs.azureSession.credentials,
-            this.inputs.azureSession.environment, this.inputs.azureSession.tenantId, this.inputs.subscriptionId);
+        this.inputs.azureSession.environment, this.inputs.azureSession.tenantId, this.inputs.subscriptionId);
         telemetryHelper.setTelemetry(TelemetryKeys.resourceType, this.inputs.targetResource.resource.type);
-        telemetryHelper.setTelemetry(TelemetryKeys.resourceKind, this.inputs.targetResource.resource.kind);
-        telemetryHelper.setTelemetry(TelemetryKeys.resourceIdHash, Utilities.createSha256Hash(this.inputs.targetResource.resource.id));
+        // Commenting below lines as for AKS these values are not set
+        // telemetryHelper.setTelemetry(TelemetryKeys.resourceKind, this.inputs.targetResource.resource.kind);
+        // telemetryHelper.setTelemetry(TelemetryKeys.resourceIdHash, Utilities.createSha256Hash(this.inputs.targetResource.resource.id));
```
bishal-pdMSFT, line 150: this should not be commented

src/configure/configure.ts
```diff
@@ -648,14 +655,54 @@ class Orchestrator {
...
+        return {
+          id: null,
+          branch,
+          pipelineTemplateId: templateId,
+          pipelineTemplateParameters: pipeplineTemplateParameters,
+          provisioningMode: mode,
+        } as ProvisioningConfiguration;
+      }
+
+    private async configurePipelineRemotely(): Promise<void>{
```
bishal-pdMSFT, line 669: create a separate configurer class for this

src/configure/configure.ts
```diff
@@ -648,14 +655,54 @@ class Orchestrator {
...
+        const provisioningConfigurer = new ProvisioningConfigurer(this.localGitRepoHelper);
+        const template = this.inputs.pipelineConfiguration.template as RemotePipelineTemplate;
+        await provisioningConfigurer.CreatePreRequisiteParams(this.inputs);
```
bishal-pdMSFT, line 672: you should log telemetry for currentstep so that we can know where failues are happening

src/configure/configure.ts
```diff
@@ -648,14 +655,54 @@ class Orchestrator {
...
+        try {
+            // Initially send the provisioning request in draft mode to get workflow files
+            const provisioningServiceDraftModeResponse  = await provisioningConfigurer.createProvisioningPipeline(provisioningConfiguration, this.inputs);
+            if (provisioningServiceDraftModeResponse.id  != "" ) {
+                const OrgAndRepoDetails =  this.inputs.sourceRepository.repositoryId.split('/');
+
+                // Monitor the provisioning pipeline
+                const provisioningDraftPipeline =  await provisioningConfigurer.checkProvisioningPipeline(provisioningServiceDraftModeResponse.id, OrgAndRepoDetails[0], OrgAndRepoDetails[1], this.inputs );
+
+                provisioningConfiguration.provisioningMode =  provisioningMode.complete;
```
bishal-pdMSFT, line 683: where are we showing draft files to user and notification to commit & push

src/configure/configure.ts
```diff
@@ -648,14 +655,54 @@ class Orchestrator {
             throw error;
         }
     }
+
+    private createProvisioningConfiguration(templateId: string, branch: string, pipeplineTemplateParameters: { [key: string]: string }, mode: provisioningMode): ProvisioningConfiguration {
```
bishal-pdMSFT, line 659: `createProvisioningConfigurationObject`

src/configure/configure.ts
bishal-pdMSFT: actually it is okay

src/configure/configurers/provisioningConfigurer.ts
```diff
@@ -0,0 +1,218 @@
...
+                    telemetryHelper.setTelemetry(TelemetryKeys.BrowsePipelineClicked, 'true');
+                    vscode.env.openExternal(vscode.Uri.parse(this.queuedPipelineUrl));
+                }
+            });
+    }
+
+    public async postSteps(provisioningConfiguration: ProvisioningConfiguration, draftPipelineConfiguration: DraftPipelineConfiguration, inputs: WizardInputs): Promise<void> {
+        await this.getFileToCommit(draftPipelineConfiguration);
+        await this.showPipelineFiles();
+        const displayMessage = (this.filesToCommit.length > 1) ?  Messages.modifyAndCommitMultipleFiles : Messages.modifyAndCommitFile;
```
bishal-pdMSFT, line 91: this should be done by caller. See how it was being done currently. it also should set proper currentstep

src/configure/configurers/provisioningConfigurer.ts
```diff
@@ -0,0 +1,218 @@
...
+                    } else {
+                        throw new Error("Failed to configure pipeline");
+                    }
+                } catch (error) {
+                    telemetryHelper.logError(Layer, TracePoints.RemotePipelineConfiguringFailed, error);
+                    vscode.window.showErrorMessage(utils.format(Messages.ConfiguringPipelineFailed, error.message));
+                    return null;
+                }
+            });
+            } else {
```
bishal-pdMSFT, line 114: the indentation is not proper

src/configure/configurers/provisioningConfigurer.ts
```diff
@@ -0,0 +1,218 @@
...
+                throw error;
+            }
+        });
+        telemetryHelper.setTelemetry(TelemetryKeys.WorkflowFileName, fileName);
+        return Path.join(directoryPath, fileName);
+    }
+
+    public async CreatePreRequisiteParams(wizardInputs: WizardInputs): Promise<void>{
+        const parsedCredentials = JSON.parse(JSON.stringify(wizardInputs.azureSession.credentials));
+        wizardInputs.pipelineConfiguration.params["armAuthToken"] = "Bearer " + parsedCredentials["tokenCache"]["target"]["_entries"][0]["accessToken"];
```
bishal-pdMSFT, line 170: Is this token safe to take dependency on? two aspects:
1. This looks like an internal detail for them. This can be prone to change/break.
2. What is the interval of refresh happening on this cache?

Should we fetch ARM token from AAD and then use that?

bishal-pdMSFT: /cc @anuragc617

src/configure/configurers/provisioningConfigurer.ts
```diff
@@ -0,0 +1,218 @@
...
+        const template = wizardInputs.pipelineConfiguration.template as RemotePipelineTemplate;
+        const inputDescriptor = template.parameters.inputs.find((value) => (value.type === InputDataType.Authorization && value.id === "azureAuth"));
```
bishal-pdMSFT, line 172: make this function more generic by adding a switch case and just calling separate functions for each different type of input. And move handling azureAuth input type to a separate function.

src/configure/helper/LocalGitRepoHelper.ts
```diff
@@ -197,4 +197,10 @@ export class LocalGitRepoHelper {
             manifestFilePath);
         await vscode.window.showTextDocument(vscode.Uri.file(manifestFilePath));
     }
+
+    public async readFileConetent(pathToFile: string): Promise<string>{
```
bishal-pdMSFT, line 201: typo

src/configure/helper/LocalGitRepoHelper.ts
```diff
@@ -197,4 +197,10 @@ export class LocalGitRepoHelper {
...
+        await vscode.workspace.saveAll(true);
```
bishal-pdMSFT, line 202: why doing the extra save here? If user has saved, we should take that.

bishal-pdMSFT:
One uber comment which I have called out - in the main configure function, you should enumerate all major steps and then specific implementation of that be in provisioningConfigurer.

src/configure/configure.ts
tiwarishub: In postStep function, we are showing draft files and notifying user for to modify and commit
https://github.com/microsoft/vscode-deploy-azure/pull/154/files#diff-df72344d1fd58d154bf36fc7942641e5R88

src/configure/helper/LocalGitRepoHelper.ts
tiwarishub: There can be scenario, where user just clicked on Modify and Push button without saving the file. So we are saving it and then sending to provisioning service.

src/configure/configurers/remoteConfigurer.ts
tiwarishub: done

src/configure/configure.ts
tiwarishub: done. Added If condition

src/configure/configure.ts
tiwarishub: done

src/configure/configure.ts
tiwarishub: done

src/configure/configurers/provisioningConfigurer.ts
tiwarishub: done

src/configure/configurers/provisioningConfigurer.ts
tiwarishub: Currently also we are doing this in configurer https://github.com/microsoft/vscode-deploy-azure/blob/6df6bc5b56db47f932259832c00fbcc8d38e262b/src/configure/configurers/localGithubWorkflowConfigurer.ts#L201

src/configure/configurers/provisioningConfigurer.ts
tiwarishub: got some clue from graph helper https://github.com/microsoft/vscode-deploy-azure/blob/master/src/configure/helper/graphHelper.ts#L96 . Will try this in next PR

src/configure/configure.ts
```diff
@@ -137,19 +143,21 @@ class Orchestrator {
...
         this.inputs.targetResource.resource = await azureResourceSelector.getAzureResource(this.inputs);
         this.azureResourceClient = ResourceSelectorFactory.getAzureResourceClient(targetType, this.inputs.azureSession.credentials,
-            this.inputs.azureSession.environment, this.inputs.azureSession.tenantId, this.inputs.subscriptionId);
+        this.inputs.azureSession.environment, this.inputs.azureSession.tenantId, this.inputs.subscriptionId);
         telemetryHelper.setTelemetry(TelemetryKeys.resourceType, this.inputs.targetResource.resource.type);
-        telemetryHelper.setTelemetry(TelemetryKeys.resourceKind, this.inputs.targetResource.resource.kind);
-        telemetryHelper.setTelemetry(TelemetryKeys.resourceIdHash, Utilities.createSha256Hash(this.inputs.targetResource.resource.id));
+        if ( targetType === TargetResourceType.WebApp) {
+            telemetryHelper.setTelemetry(TelemetryKeys.resourceKind, this.inputs.targetResource.resource.kind);
+            telemetryHelper.setTelemetry(TelemetryKeys.resourceIdHash, Utilities.createSha256Hash(this.inputs.targetResource.resource.id));
```
bishal-pdMSFT, line 150: Are we logging telemetry like this in case of remote provisioning? e.g. resourceId hash

src/configure/configure.ts
```diff
@@ -100,22 +103,25 @@ class Orchestrator {
...
         if (this.continueOrchestration) {
-            let pipelineConfigurer = ConfigurerFactory.GetConfigurer(this.inputs.sourceRepository, this.inputs.azureSession,
-                this.inputs.pipelineConfiguration.template.templateType, this.localGitRepoHelper);
-            let selectedCICDProvider = (pipelineConfigurer.constructor.name === "AzurePipelineConfigurer") ? constants.azurePipeline : constants.githubWorkflow;
+            if (this.inputs.sourceRepository.repositoryProvider === RepositoryProvider.Github &&
+                this.inputs.targetResource.resource.type == TargetResourceType.AKS ) {
+               return await this.configurePipelineRemotely();
+            }
+
+            const pipelineConfigurer = ConfigurerFactory.GetConfigurer(this.inputs.sourceRepository, this.inputs.azureSession,
```
bishal-pdMSFT, line 112: move this part to a separate function called `configurePipelineLocally` to keep it at par with configurePipelineRemotely

src/configure/configure.ts
```diff
@@ -137,19 +143,21 @@ class Orchestrator {
...
+            telemetryHelper.setTelemetry(TelemetryKeys.resourceIdHash, Utilities.createSha256Hash(this.inputs.targetResource.resource.id));
+        }
     }
 
     private async selectTemplate(resource: GenericResource): Promise<void> {
         switch (resource.type) {
             case TargetResourceType.AKS:
-                this.inputs.pipelineConfiguration.template = this.inputs.potentialTemplates.find((template) => template.templateType === TemplateType.LOCAL);
-                break;
+               this.inputs.pipelineConfiguration.template = this.inputs.potentialTemplates.find((template) => template.templateType === TemplateType.REMOTE);
```
bishal-pdMSFT, line 157: Please check the count of templates first and if no templates found, then log an error and abort. See L168

src/configure/configure.ts
```diff
@@ -100,22 +103,25 @@ class Orchestrator {
...
         telemetryHelper.setCurrentStep('GetAllRequiredInputs');
         await this.getInputs(node);
-
         if (this.continueOrchestration) {
-            let pipelineConfigurer = ConfigurerFactory.GetConfigurer(this.inputs.sourceRepository, this.inputs.azureSession,
-                this.inputs.pipelineConfiguration.template.templateType, this.localGitRepoHelper);
-            let selectedCICDProvider = (pipelineConfigurer.constructor.name === "AzurePipelineConfigurer") ? constants.azurePipeline : constants.githubWorkflow;
+            if (this.inputs.sourceRepository.repositoryProvider === RepositoryProvider.Github &&
+                this.inputs.targetResource.resource.type == TargetResourceType.AKS ) {
+               return await this.configurePipelineRemotely();
```
bishal-pdMSFT, line 109: This new feature should be behind a feature flag so that we can turn it off immediately if users report major issues with it

bishal-pdMSFT: For reference see `templateServiceEnabled` in `ExtensionVariables` class

src/configure/configure.ts
```diff
@@ -648,14 +656,69 @@ class Orchestrator {
...
+            // Draft pipeline step
+            telemetryHelper.setCurrentStep('ConfiguringDraftPipeline');
+            let provisioningServiceDraftModeResponse: ProvisioningConfiguration;
+            let provisioningDraftPipeline: ProvisioningConfiguration;
+            const provisioningConfiguration =  this.createProvisioningConfigurationObject(template.id, this.inputs.sourceRepository.branch, this.inputs.pipelineConfiguration.params , provisioningMode.draft );
+            await vscode.window.withProgress({ location: vscode.ProgressLocation.Notification, title: Messages.ConfiguringDraftPipeline }, async () =>
+                {
+                    try {
+                            // Initially send the provisioning request in draft mode to get workflow files
+                            provisioningServiceDraftModeResponse  = await provisioningConfigurer.createProvisioningPipeline(provisioningConfiguration, this.inputs);
```
bishal-pdMSFT, line 687: `queueProvisioningPipelineJob`

src/configure/configure.ts
```diff
@@ -648,14 +656,69 @@ class Orchestrator {
...
+                            if (provisioningServiceDraftModeResponse.id  != "" ) {
+                                // Monitor the provisioning pipeline
+                                const OrgAndRepoDetails =  this.inputs.sourceRepository.repositoryId.split('/');
+                                provisioningDraftPipeline =  await provisioningConfigurer.checkProvisioningPipeline(provisioningServiceDraftModeResponse.id, OrgAndRepoDetails[0], OrgAndRepoDetails[1], this.inputs );
```
bishal-pdMSFT, line 691: `awaitProvisioningPipelineJob`

src/configure/configure.ts
```diff
@@ -648,14 +656,69 @@ class Orchestrator {
...
+        const provisioningConfigurer = new ProvisioningConfigurer(this.localGitRepoHelper);
+        const template = this.inputs.pipelineConfiguration.template as RemotePipelineTemplate;
+        try {
+            // prerequisite params
+            telemetryHelper.setCurrentStep('ConfiguringPreRequisiteParams');
+            await provisioningConfigurer.CreatePreRequisiteParams(this.inputs);
+
+            // Draft pipeline step
+            telemetryHelper.setCurrentStep('ConfiguringDraftPipeline');
+            let provisioningServiceDraftModeResponse: ProvisioningConfiguration;
```
bishal-pdMSFT, line 680: should we move this job queuing and waiting part to the configurer? This way the current function can be simpler.

bishal-pdMSFT: A counterargument could be that we want to treat queueing and waiting as two separate steps. In that case, you can keep the logic here, but then you should log setCurrentStep for both separately.

src/configure/configure.ts
```diff
@@ -648,14 +656,69 @@ class Orchestrator {
...
+            let provisioningDraftPipeline: ProvisioningConfiguration;
```
bishal-pdMSFT, line 681: `draftProvisioningPipeline`

src/configure/configurers/provisioningConfigurer.ts
```diff
@@ -0,0 +1,267 @@
...
+                if ( provisioningServiceResponse.result.status ===  "Queued" ||  provisioningServiceResponse.result.status == "InProgress") {
+                await sleepForMilliSeconds(this.refreshTime);
+                return await this.checkProvisioningPipeline(jobId, githubOrg, repository, wizardInputs);
+                } else if (provisioningServiceResponse.result.status ===  "Failed") {
+                throw new Error(provisioningServiceResponse.result.message) ;
+                } else {
+                    return provisioningServiceResponse;
+                }
+            }else {
+                throw new Error("Failed to receive queued pipeline provisioning job status");
```
bishal-pdMSFT, line 77: we should not fail in case we fail to get a result. Rather continue and proceed to next get.

bishal-pdMSFT: But yes, we will have to keep a max number of such tries so that we do not keep re-trying which just keeps failing.

src/configure/configurers/provisioningConfigurer.ts
bishal-pdMSFT: ok

src/configure/configurers/provisioningConfigurer.ts
```diff
@@ -0,0 +1,267 @@
...
+        dirList.forEach((dir) => {
+            try {
+                directoryPath = Path.join(directoryPath, dir);
+                // tslint:disable-next-line:non-literal-fs-path
+                if (!fs.existsSync(directoryPath)) {
+                    // tslint:disable-next-line:non-literal-fs-path
+                    fs.mkdirSync(directoryPath);
+                }
+            }
+            catch (error) {
```
bishal-pdMSFT, line 168: you don't need to catch error just to throw it again. If you don't write a catch block, the error will be automatically thrown to parent

src/configure/configurers/provisioningConfigurer.ts
```diff
@@ -0,0 +1,267 @@
...
+        // Create SPN and ACRResource group for reuseACR flow set to false
+        const inputDescriptor = this.getInputDescriptor(wizardInputs, "azureAuth");
+        if (inputDescriptor != undefined){
+            const createResourceGroup = InputControl.getInputDescriptorProperty(inputDescriptor, "createResourceGroup", wizardInputs.pipelineConfiguration.params);
+            if ( createResourceGroup.length > 0 && createResourceGroup[0] != "" ){
+                await vscode.window.withProgress(
+                    { location: vscode.ProgressLocation.Notification, title: Messages.CreatingACRResourceGroup },
+                    async () => {
+                        try {
+                           return await new ArmRestClient(wizardInputs.azureSession).createResourceGroup(wizardInputs.subscriptionId, createResourceGroup[0], wizardInputs.pipelineConfiguration.params["containerRegistryLocation"]);
```
bishal-pdMSFT, line 186: Okay we need location as well, and I think it is wrong to take dependency on containerRegistryLocation as we might be creating RG for some other purpose in future. Should we take an array of  <rg name, location> instead of rg name?

bishal-pdMSFT: /cc @anuragc617

src/configure/helper/LocalGitRepoHelper.ts
bishal-pdMSFT: @tiwarishub when we display draft files to user, we would have already saved the files. is that right? if so, then we will let user save the changes he made, otherwise just commit the files which we saved earlier.

src/configure/configurers/provisioningConfigurer.ts
tiwarishub: kept try and cath and added logging in catch block

src/configure/configurers/provisioningConfigurer.ts
tiwarishub: Added changes to take location from template. Similar changes are at template service as well.

src/configure/configurers/provisioningConfigurer.ts
tiwarishub: I have a counter which run valid for 5 mins. after that we will throw error

src/configure/configure.ts
tiwarishub: added

src/configure/configure.ts
tiwarishub: Added two different. for Complete case we are logging steps in postSteps function

src/configure/configure.ts
tiwarishub: done

src/configure/configure.ts
tiwarishub: done

src/configure/configure.ts
tiwarishub: Added feature flag deployToAzure.UseRemoteConfigurer

src/configure/configure.ts
tiwarishub: Added check, if find is getting failed  for remote templates then we will through error

src/configure/configure.ts
tiwarishub: done

src/configure/configure.ts
tiwarishub: This seems a bug to me for AKS case, because for AKS these values will be undefined  at this point of function this.inputs.targetResource.resource.kind,
this.inputs.targetResource.resource.id.

tiwarishub: Okay, got it. You mean we need to log hash of aks cluster id as well. Currently we are not doing. I will add it

tiwarishub: Will change this as a part of template service template to make it generic. there we have params logtelemetry and hash. that we can use

package.json
```diff
@@ -103,6 +103,11 @@
                     "type": "boolean",
                     "default": true,
                     "description": "Use GitHub for creating new repository"
+                },
+                "deployToAzure.UseRemoteConfigurer": {
```
bishal-pdMSFT, line 107: I don't think we should have this as a configuration setting? It is confusing for user to understand this variable.
We should just use a global variable for this. If customers report problem, we will set this variable to false and release a new version of vs code.

src/configure/configure.ts
bishal-pdMSFT: makes sense

package.json
tiwarishub: Okay, I will remove it from here and keep in extension variables and set to value to true

src/configure/configure.ts
bishal-pdMSFT: Nice

bishal-pdMSFT: You can still move this code till L687 to the configurer to keep current function small.

src/configure/configure.ts
```diff
@@ -653,14 +643,100 @@ class Orchestrator {
...
+                    try {
+                            // Initially send the provisioning request in draft mode to get workflow files
+                            provisioningServiceDraftModeResponse  = await provisioningConfigurer.queueProvisioningPipelineJob(provisioningConfiguration, this.inputs);
+                            if (provisioningServiceDraftModeResponse.id  != "" ) {
+                                // Monitor the provisioning pipeline
+                                telemetryHelper.setCurrentStep('AwaitDraftProvisioningPipeline');
+                                const OrgAndRepoDetails =  this.inputs.sourceRepository.repositoryId.split('/');
+                                draftProvisioningPipeline =  await provisioningConfigurer.awaitProvisioningPipelineJob(provisioningServiceDraftModeResponse.id, OrgAndRepoDetails[0], OrgAndRepoDetails[1], this.inputs );
+                            }else {
+                                throw new Error("Failed to configure provisioning pipeline");
```
bishal-pdMSFT, line 681: we should throw the error returned by the API

tiwarishub: We are throwing the same error returned by API. This else condition is for corner case where the return repsonose code is 2XX but we haven't received job id. In this case we are making an error

src/configure/configurers/provisioningConfigurer.ts
```diff
@@ -0,0 +1,280 @@
...
+                } else {
+                    return provisioningServiceResponse;
+                }
+            }else {
+                if ( statusNotFound < this.maxNonStatusRetry ) {
+                    statusNotFound++;
+                    await sleepForMilliSeconds(this.refreshTime);
+                    return await this.awaitProvisioningPipelineJob(jobId, githubOrg, repository, wizardInputs);
+                } else {
+                 throw new Error("Failed to receive queued pipeline provisioning job status");
```
bishal-pdMSFT, line 84: we should throw the error returned by the API. Please do so other places as well.

src/configure/configurers/provisioningConfigurer.ts
```diff
@@ -0,0 +1,280 @@
...
+        return await this.provisioningServiceClient.getProvisioningConfiguration(jobId, githubOrg, repository);
+       } catch (error){
+            telemetryHelper.logError(Layer, TracePoints.UnabletoGetProvisioningPipeline, error);
+            throw error;
+       }
+    }
+
+    public async awaitProvisioningPipelineJob(jobId: string, githubOrg: string, repository: string, wizardInputs: WizardInputs): Promise<ProvisioningConfiguration> {
+        let statusNotFound: number = 0;
+        try {
```
bishal-pdMSFT, line 67: I think we are already handling all scenarios by looking at API responses. We do not need this separate try-catch.

src/configure/configurers/provisioningConfigurer.ts
tiwarishub: This is also for same case. Return response is 2XX , but result not received in response.

src/configure/configurers/provisioningConfigurer.ts
tiwarishub: Okay i will change this

src/configure/configurers/provisioningConfigurer.ts
```diff
@@ -0,0 +1,280 @@
...
+    constructor(localGitRepoHelper: LocalGitRepoHelper){
+        this.localGitRepoHelper = localGitRepoHelper;
+    }
+
+    public async queueProvisioningPipelineJob(provisioningConfiguration: ProvisioningConfiguration, wizardInputs: WizardInputs): Promise<ProvisioningConfiguration>{
+        try {
+            this.provisioningServiceClient =  await ProvisioningServiceClientFactory.getClient(wizardInputs.githubPATToken, wizardInputs.azureSession.credentials);
+            const OrgAndRepoDetails = wizardInputs.sourceRepository.repositoryId.split('/');
+            return await this.provisioningServiceClient.createProvisioningConfiguration(provisioningConfiguration, OrgAndRepoDetails[0], OrgAndRepoDetails[1]);
+        } catch (error){
```
bishal-pdMSFT, line 49: I am seeing too many try-catch blocks. We should only have try-catch when we want to catch and log telemetry. I don't think we need separate telemetry for all failures. We should have telemetry at a bit higher level. For example, it can be in configurer.

bishal-pdMSFT: Ah this is configurer only. So it is okay.

src/configure/configurers/provisioningConfigurer.ts
```diff
@@ -0,0 +1,280 @@
...
+                try {
+                    telemetryHelper.setCurrentStep('QueuedCompleteProvisioiningPipeline');
+                    provisioningConfiguration.pipelineConfiguration = await this.createFilesToCheckin(draftPipelineConfiguration.id, draftPipelineConfiguration.type);
+                    const completeProvisioningSvcResp = await this.queueProvisioningPipelineJob(provisioningConfiguration, inputs);
+                    if ( completeProvisioningSvcResp.id != "" ){
+                        const OrgAndRepoDetails = inputs.sourceRepository.repositoryId.split('/');
+                        telemetryHelper.setCurrentStep('AwaitCompleteProvisioningPipeline');
+                        return await this.awaitProvisioningPipelineJob(completeProvisioningSvcResp.id, OrgAndRepoDetails[0], OrgAndRepoDetails[1], inputs);
+                    } else {
+                        throw new Error("Failed to configure pipeline");
```
bishal-pdMSFT, line 124: throw the error returned by API

src/configure/configure.ts
bishal-pdMSFT: got it.

src/configure/configurers/provisioningConfigurer.ts
tiwarishub: this is for same reason, response is 2xx. But payload doesn't have job id. So we are throwing our own error

src/configure/configurers/provisioningConfigurer.ts
```diff
@@ -0,0 +1,280 @@
...
+            telemetryHelper.logError(Layer, TracePoints.UnabletoGetProvisioningPipeline, error);
+            throw error;
+       }
+    }
+
+    public async awaitProvisioningPipelineJob(jobId: string, githubOrg: string, repository: string, wizardInputs: WizardInputs): Promise<ProvisioningConfiguration> {
+        let statusNotFound: number = 0;
+        try {
+            const provisioningServiceResponse = await this.getProvisioningPipeline(jobId, githubOrg, repository, wizardInputs);
+            if ( provisioningServiceResponse.result){
```
bishal-pdMSFT, line 69: do we need to check for `if ( provisioningServiceResponse && provisioningServiceResponse.result)`

src/configure/configurers/provisioningConfigurer.ts
```diff
@@ -0,0 +1,280 @@
...
+                if ( provisioningServiceResponse.result.status ===  "Queued" ||  provisioningServiceResponse.result.status == "InProgress") {
+                await sleepForMilliSeconds(this.refreshTime);
```
bishal-pdMSFT, line 71: fix formatting

src/configure/configurers/provisioningConfigurer.ts
```diff
@@ -0,0 +1,280 @@
...
+                return await this.awaitProvisioningPipelineJob(jobId, githubOrg, repository, wizardInputs);
+                } else if (provisioningServiceResponse.result.status ===  "Failed") {
+                throw new Error(provisioningServiceResponse.result.message) ;
+                } else {
+                    return provisioningServiceResponse;
+                }
+            }else {
+                if ( statusNotFound < this.maxNonStatusRetry ) {
```
bishal-pdMSFT, line 79: so, this will get executed when `provisioningServiceResponse` is present but `provisioningServiceResponse.Result` is null. What case is that?
Should we re-try if any sort of failure happens in `getProvisioningPipeline`?

bishal-pdMSFT (approved):

src/configure/configurers/provisioningConfigurer.ts
bishal-pdMSFT: makes sense

src/configure/configurers/provisioningConfigurer.ts
bishal-pdMSFT: ok

src/configure/configurers/provisioningConfigurer.ts
tiwarishub: Yes, this can happen if at provisioning service before sending response we are not properly converting to contracts or result  (generally i don't think this will happen, as at provisioning service we have proper conversion).
For re-try in `getProvisioningPipeline`, this function is quite simple and for failure in http request we have retry logic in ms-rest client.

src/configure/configurers/provisioningConfigurer.ts
tiwarishub: done

src/configure/configurers/provisioningConfigurer.ts
tiwarishub: added this

src/configure/configurers/provisioningConfigurer.ts
tiwarishub: done

package.json
tiwarishub: done

src/configure/configure.ts
tiwarishub: Thanks for suggesting this changes. This change has made this function quite simple.

PR #163, opened 2020-09-22
Acquire token for ArmAuthtoken

tiwarishub:
removed the previous logic to get token


bishal-pdMSFT (approved):

src/configure/helper/graphHelper.ts
```diff
@@ -87,6 +83,14 @@ export class GraphHelper {
 
         return accountName + "-" + projectName + "-" + guid;
     }
+    public static async getAccessToken(session: AzureSession): Promise<string>{
+      const token = await this.getRefreshToken(session);
+      return token.accessToken;
+     }
+
+    private static contributorRoleId = "b24988ac-6180-42a0-ab88-20f7382dd24c";
```
anuragc617, line 91: data members should come in the beginning of a class.

src/configure/helper/graphHelper.ts
```diff
@@ -87,6 +83,14 @@ export class GraphHelper {
 
         return accountName + "-" + projectName + "-" + guid;
     }
+    public static async getAccessToken(session: AzureSession): Promise<string>{
```
anuragc617, line 86: nit: new line

anuragc617 (approved):

src/configure/helper/graphHelper.ts
tiwarishub: yes, that correct but private data members should come after public data members thats i moved them after public functions

PR #165, opened 2020-09-24
AKS right click scenario support - Configure CI/CD

kanika1894:
This will work in complement to the right click scenario support in AKS extension.

Templates are filtered on resource type i.e. AKS.

If the cluster is selected by right click, cluster details are stored in context, which are used while gettings inputs as per the template from template service.

### DIFF
--- a/src/configure/configure.ts
+++ b/src/configure/configure.ts
@@ -186,15 +186,6 @@ class Orchestrator {
 
                 if (this.inputs.targetResource.resource && this.inputs.targetResource.resource.id) {
                     this.context['resourceId'] = this.inputs.targetResource.resource.id;
-                    // Getting cluster related properties for AKS scenario
-                    if (this.inputs.targetResource.resource.type === TargetResourceType.AKS) {
-                        this.context['kubernetesVersion'] = this.inputs.targetResource.resource.properties.kubernetesVersion;
-                        const addonProfiles = JSON.parse(JSON.stringify(this.inputs.targetResource.resource.properties.addonProfiles).toLowerCase());
-                        this.context['httpApplicationRoutingEnabled'] = addonProfiles.httpapplicationrouting.enabled;
-                        if (this.context['httpApplicationRoutingEnabled']) {
-                            this.context['httpApplicationRoutingDomain'] = addonProfiles.httpapplicationrouting.config.httpapplicationroutingzonename;
-                        }
-                    }
                 }
 
                 this.selectTemplate(this.inputs.targetResource.resource);
--- a/src/configure/configure.ts
+++ b/src/configure/configure.ts
@@ -128,7 +128,7 @@ class Orchestrator {
     private async selectTemplate(resource: GenericResource): Promise<void> {
         switch (resource.type) {
             case TargetResourceType.AKS:
-                if ( extensionVariables.remoteConfigurerEnabled === true && this.inputs.sourceRepository.repositoryProvider === RepositoryProvider.Github ) {
+                if (extensionVariables.remoteConfigurerEnabled === true && this.inputs.sourceRepository.repositoryProvider === RepositoryProvider.Github) {
                     this.inputs.pipelineConfiguration.template = this.inputs.potentialTemplates.find((template) => template.templateType === TemplateType.REMOTE);
                 } else {
                     this.inputs.pipelineConfiguration.template = this.inputs.potentialTemplates.find((template) => template.templateType === TemplateType.LOCAL);
@@ -176,6 +176,7 @@ class Orchestrator {
 
             try {
                 if (!resourceNode) {
+                    this.context['resourceId'] = '';
                     this.context['isResourceAlreadySelected'] = false;
                     await this.getAzureSubscription();
                     await this.getAzureResource(this.getSelectedPipelineTargetType());
--- a/src/configure/configure.ts
+++ b/src/configure/configure.ts
@@ -104,11 +104,10 @@ class Orchestrator {
         telemetryHelper.setCurrentStep('GetAllRequiredInputs');
         await this.getInputs(node);
         if (this.continueOrchestration) {
-            if (extensionVariables.remoteConfigurerEnabled === true && this.inputs.sourceRepository.repositoryProvider === RepositoryProvider.Github &&
-                this.inputs.targetResource.resource.type === TargetResourceType.AKS) {
-                return await this.configurePipelineRemotely();
+            if ( extensionVariables.remoteConfigurerEnabled === true && this.inputs.sourceRepository.repositoryProvider === RepositoryProvider.Github &&
+                this.inputs.targetResource.resource.type === TargetResourceType.AKS && !!this.inputs.sourceRepository.remoteUrl ) {
+               return await this.configurePipelineRemotely();
             }
-
             return this.ConfigurePipelineLocally();
         }
     }
@@ -128,7 +127,7 @@ class Orchestrator {
     private async selectTemplate(resource: GenericResource): Promise<void> {
         switch (resource.type) {
             case TargetResourceType.AKS:
-                if (extensionVariables.remoteConfigurerEnabled === true && this.inputs.sourceRepository.repositoryProvider === RepositoryProvider.Github) {
+                if ( extensionVariables.remoteConfigurerEnabled === true && this.inputs.sourceRepository.repositoryProvider === RepositoryProvider.Github && !!this.inputs.sourceRepository.remoteUrl) {
                     this.inputs.pipelineConfiguration.template = this.inputs.potentialTemplates.find((template) => template.templateType === TemplateType.REMOTE);
                 } else {
                     this.inputs.pipelineConfiguration.template = this.inputs.potentialTemplates.find((template) => template.templateType === TemplateType.LOCAL);
--- a/src/configure/configure.ts
+++ b/src/configure/configure.ts
@@ -186,12 +186,13 @@ class Orchestrator {
 
                 if (this.inputs.targetResource.resource && this.inputs.targetResource.resource.id) {
                     this.context['resourceId'] = this.inputs.targetResource.resource.id;
+                    // Getting cluster related properties for AKS scenario
                     if (this.inputs.targetResource.resource.type === TargetResourceType.AKS) {
                         this.context['kubernetesVersion'] = this.inputs.targetResource.resource.properties.kubernetesVersion;
-                        let addonProfiles = JSON.parse(JSON.stringify(this.inputs.targetResource.resource.properties.addonProfiles).toLowerCase());
+                        const addonProfiles = JSON.parse(JSON.stringify(this.inputs.targetResource.resource.properties.addonProfiles).toLowerCase());
                         this.context['httpApplicationRoutingEnabled'] = addonProfiles.httpapplicationrouting.enabled;
                         if (this.context['httpApplicationRoutingEnabled']) {
-                            this.context['httpApplicationRoutingDomain'] = addonProfiles.httpapplicationrouting.httpapplicationroutingZoneName;
+                            this.context['httpApplicationRoutingDomain'] = addonProfiles.httpapplicationrouting.config.httpapplicationroutingzonename;
                         }
                     }
                 }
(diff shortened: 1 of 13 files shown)


src/configure/configure.ts
```diff
@@ -186,7 +186,17 @@ class Orchestrator {
 
                 if (this.inputs.targetResource.resource && this.inputs.targetResource.resource.id) {
                     this.context['resourceId'] = this.inputs.targetResource.resource.id;
+                    // Getting cluster related properties for AKS scenario
+                    if (this.inputs.targetResource.resource.type === TargetResourceType.AKS) {
```
anuragc617, line 190: this is specific to AKS can be make it generic to get the various properties of a resource.

src/configure/configure.ts
```diff
@@ -254,7 +264,7 @@ class Orchestrator {
     }
 
     private async analyzeNode(node: any): Promise<GenericResource> {
-        if (!!node && !!node.fullId) {
+        if (!!node && !!node.cloudResource.id) {
```
anuragc617, line 267: can it break the existing right click scenarios for webApp ??

src/configure/configure.ts
```diff
@@ -430,7 +440,6 @@ class Orchestrator {
     }
 
     private async extractAzureResourceFromNode(node: AzureTreeItem | any): Promise<GenericResource> {
-        const resource: GenericResource = null;
         if (!!node.fullId) {
```
anuragc617, line 443: here you are checking for fullId and in line 267 you have changed this condition

src/configure/templateInputHelper/InputControlProvider.ts
```diff
@@ -192,7 +192,7 @@ export class InputControlProvider {
 
         for (var clientInput of uniqueDependentClientInputs) {
             var dependentInputControl = this._context[clientInput];
-            if (dependentInputControl) {
+            if (dependentInputControl !== undefined) {
```
anuragc617, line 195: why this is required.

kanika1894: If some value is set to be 'false' in context, this condition evaluates it incorrectly and throws the error that the value doesn't exist in context, though it exists.

src/configure/configure.ts
kanika1894: The AKS specific work was thee from line 458, where I have updated the condition
```         else if (!!node.cloudResource && node.cloudResource.nodeType === 'cluster') ```

Didn't change this at 434 as not making any difference to us now, and would be better to update if when we are integrating webApp right click scenario for which the code seems to be written.

File: src/configure/configure.ts (hunk 2 above)
Comment (kanika1894) [reply src/configure/configure.ts:258]: Is the right click scenario for web app fully implemented?
I thought that is still pending, and this thing will be verified when that is being implemented.

Comment (bishal-pdMSFT) [reply src/configure/configure.ts:258]: Agree. `cloudResource` property might not exist for web app. Better to make this check for AKS case only in extractAzureResourceFromNode

File: src/configure/configure.ts [hunk 5]
```diff
@@ -254,7 +255,7 @@ class Orchestrator {
     }
 
     private async analyzeNode(node: any): Promise<GenericResource> {
-        if (!!node && !!node.fullId) {
+        if (!!node && !!node.cloudResource.id) {
```
bishal-pdMSFT, line 258: Does AKS extension publish any type definition for cluster node type? If so, may be we should use that.

src/configure/configure.ts
```diff
@@ -455,27 +455,29 @@ class Orchestrator {
...
-            this.inputs.subscriptionId = node.value.subscription.subscriptionId;
+        else if (!!node.cloudResource && node.cloudResource.nodeType === 'cluster') {
+            this.inputs.subscriptionId = node.cloudResource.subscription.subscriptionId;
+            this.context['subscriptionId'] = this.inputs.subscriptionId;
             this.inputs.azureSession = getSubscriptionSession(this.inputs.subscriptionId);
             this.azureResourceClient = new AzureResourceClient(this.inputs.azureSession.credentials, this.inputs.subscriptionId);
-            const cluster = await this.azureResourceClient.getResource(node.value.armId, '2019-08-01');
+            const cluster = await this.azureResourceClient.getResource(node.cloudResource.id, '2019-08-01');
             telemetryHelper.setTelemetry(TelemetryKeys.resourceType, cluster.type);
             telemetryHelper.setTelemetry(TelemetryKeys.resourceKind, cluster.kind);
```
bishal-pdMSFT, line 465: what does `kind` represent here?

kanika1894: I don't think "kind" should exist for cluster type resource. Will verify, and remove.

src/configure/configure.ts
```diff
@@ -512,13 +514,13 @@ class Orchestrator {
             appropriatePipelines = remotePipelines;
         }
         const pipelineMap = this.getMapOfUniqueLabels(appropriatePipelines);
+        this.filterPipelinesByResourceType(pipelineMap);
```
bishal-pdMSFT, line 517: We are still getting list of all applicable templates and then filtering here. I think that is redundant. We should get list of templates.
We need to break repo analysis and listing templates in two parts, and then do only repo analysis in this case.

bishal-pdMSFT, line 517: Rather than filtering these templates, I think it would be better to call the other API which returns one template based on language + build target + deploy target.
@anuragc617 any thoughts here?

kanika1894: Yes, we are getting all templatesand then filtering.
for example, for AKS scenario itself, there can be 2 templates - helm chart and otherwise. So, in that scenario, we'll just filter on the basis of target, and will let the user decide which path to go.

bishal-pdMSFT, line 517: What should happen if dockerfile does not exist? I think we should show an error to user and tell them to add a dockerfile and try again. This should happen immediately after repo analysis.

bishal-pdMSFT: Ah, makes sense. Then my other comment about using a different template service API is redundant

kanika1894: As of now, if there isn't any dockerFil, this would mean that the template service will not return AKS template and after the filterByResource, the templates to proceed will be 0 which will give the error that the language of the repo is not supported.

bishal-pdMSFT: So, based on your [comment](https://github.com/microsoft/vscode-deploy-azure/pull/165/files#r494124525), this comment will change to: what should happen if returned list does contain any template with target of AKS?

kanika1894: So, the templates with AKS target are filtered. If they are more than one, user gets to choose, otherwise, considering present scenario, if there is one, we proceed as it is. This is in sync with the existing behavior that we have.

src/configure/configure.ts
```diff
@@ -533,6 +535,16 @@ class Orchestrator {
         }
     }
 
+    private filterPipelinesByResourceType(pipelineMap: Map<string, PipelineTemplate[]>) {
+        if (this.inputs.targetResource.resource && this.inputs.targetResource.resource.id) {
+            pipelineMap.forEach((element) => {
+                if (element.length > 0 && element[0].targetType !== this.inputs.targetResource.resource.type) {
+                    pipelineMap.delete(element[0].label);
```
bishal-pdMSFT, line 542: why check for `element.length > 0`. can that even happen?

bishal-pdMSFT, line 542: I have seen deletion of an element from a map while iterating through the map causes problem in c#. is that not a problem in JS?

src/configure/configure.ts
kanika1894: They have defined their own namespace which contains the structure of CloudExplorerResourceNode which we are getting in our extension. The cloudResource is of type any.

image

src/configure/configure.ts
kanika1894: Not required

kanika1894: Tested a couple os scenarios, worked fine.
Any particular kind of issue caused in C#, which you can mention, so that I'll check that?

src/configure/configure.ts
bishal-pdMSFT: > As of now, if there isn't any dockerFile, this would mean that the template service should not return AKS template and after the filterByResource, the templates to proceed will be 0 which will give the error that the language of the repo is not supported.

can you check this behavior with Pulkit? In case dockerfile is not found, we might still want to proceed by creating a workflow by taking a default dockerfile path. This is because we know that user wants to deploy to AKS and hence we  do not fail just because dockerfile is not found. May be used can create one before they do "commit and push"

src/configure/configure.ts
bishal-pdMSFT: I think if failed in removing the element itself. So it seems it would be okay here.

src/configure/configure.ts
```diff
@@ -254,7 +255,7 @@ class Orchestrator {
     }
 
     private async analyzeNode(node: any): Promise<GenericResource> {
-        if (!!node && !!node.fullId) {
+        if (!!node) {
```
bishal-pdMSFT, line 258: this might break "right click on folder" as node will not be null, but no azure resource will be found. I think you should just call `this.extractAzureResourceFromNode(node)` always and let that function return true/false depending on whether it found a resource. If false, then go to L262.

src/configure/configure.ts
bishal-pdMSFT: I was discussing this with @anuragc617 and he mentioned that if we use the template service filter API then it will return all AKS specific templates. This way we won't have a hard dependency on repo analysis. The templates will be shown to user even if they do not have dockerfile. If they select the template, then dockerfile path value will be the default value specified in template. This way we won't block user from creating workflow even if they don't have dockerfile. This behavior will then be consistent with AKS deployment center experience.

bishal-pdMSFT: I am okay to take this a separate PR so that we can deploy v1 first and unblock integration with AKS extension. This improvement can be taken in parallel.

bishal-pdMSFT (approved):

bishal-pdMSFT (changes requested):
please fix https://github.com/microsoft/vscode-deploy-azure/pull/165/files#r494977401

src/configure/configure.ts
kanika1894: So, just to confirm, the flow will be :
1. Calling repo analysis
2. If the reource is already selected, calling template service filter API based on the target resource, else keeping the flow as it is which gets templates based on the repo analysis response
3. Further flow remains the same

Right?
Will pick this in another PR

bishal-pdMSFT: Yes 👍

src/configure/configure.ts
```diff
@@ -255,14 +255,12 @@ class Orchestrator {
     }
 
     private async analyzeNode(node: any): Promise<GenericResource> {
-        if (!!node) {
-            return await this.extractAzureResourceFromNode(node);
-        }
-        else if (node && node.fsPath) {
-            this.workspacePath = node.fsPath;
-            telemetryHelper.setTelemetry(TelemetryKeys.SourceRepoLocation, SourceOptions.CurrentWorkspace);
+        if (!!node && !await this.extractAzureResourceFromNode(node)) {
```
bishal-pdMSFT, line 258: lets extract the filepath thing also in a similar function to make the code simpler to read.

```
if (!!node) {
    resourceNodeFound = await this.extractAzureResourceFromNode(node);
    if(!resourceNodeFound) {
       await this.extractFolderInfoFromNode(node);
    }
}
```

src/configure/configure.ts
```diff
@@ -176,17 +175,19 @@ class Orchestrator {
...
                     await this.getAzureSubscription();
                     await this.getAzureResource(this.getSelectedPipelineTargetType());
                 }
                 else {
-                    this.context['rightClickScenario'] = true;
+                    this.context['isResourceAlreadySelected'] = true;
                 }
 
                 if (this.inputs.targetResource.resource && this.inputs.targetResource.resource.id) {
                     this.context['resourceId'] = this.inputs.targetResource.resource.id;
```
bishal-pdMSFT, line 188: should we not do `this.context['isResourceAlreadySelected'] = true;` here rather than in L184. This way there is no inconsistency. We should not have isResourceAlreadySelected = true but resourceId as empty.

bishal-pdMSFT: Thinking more about it, this looks okay

bishal-pdMSFT: Or better move setting these context items to `analyzeRepo` function. Overall concern is that same context is being set at multiple places making it difficult to understand.

bishal-pdMSFT (approved):

PR #170, opened 2020-09-29
Getting templates based filtered by resource

kanika1894:
For scenarios such as right click on cluster, the performance can be improved if instead of getting all the templates and then filtering it by resource type, we can get templates based on the resource type itself.

If we already hve a resource selected, we can pass the language and deploy target to get the rquired templates.

### DIFF
--- a/src/configure/clients/ITemplateServiceClient.ts
+++ b/src/configure/clients/ITemplateServiceClient.ts
@@ -8,5 +8,5 @@ export interface ITemplateServiceClient {
     getTemplateParameters(templateId: string): Promise<ExtendedPipelineTemplate>;
     getTemplateConfiguration(templateId: string, inputs: StringMap<string>): Promise<ExtendedPipelineTemplate>;
     getTemplateFile(templateId: string, fileName: string): Promise<{ id: string, content: string }[]>;
-    getResourceFilteredTemplates(language: string, deployTarget: string): Promise<TemplateInfo[]>;
+    getTemplatesInfoByFilter(language: string, deployTargetFilter?: string, buildTargetFilter?: string): Promise<TemplateInfo[]>;
 }
--- a/src/configure/clients/ITemplateServiceClient.ts
+++ b/src/configure/clients/ITemplateServiceClient.ts
@@ -8,5 +8,5 @@ export interface ITemplateServiceClient {
     getTemplateParameters(templateId: string): Promise<ExtendedPipelineTemplate>;
     getTemplateConfiguration(templateId: string, inputs: StringMap<string>): Promise<ExtendedPipelineTemplate>;
     getTemplateFile(templateId: string, fileName: string): Promise<{ id: string, content: string }[]>;
-    getTemplatesInfoByFilter(resourceType: string): Promise<TemplateInfo[]>;
+    getTemplatesInfoByFilter(language: string, deployTargetFilter?: string, buildTargetFilter?: string): Promise<TemplateInfo[]>;
 }
--- a/src/configure/clients/ITemplateServiceClient.ts
+++ b/src/configure/clients/ITemplateServiceClient.ts
@@ -8,4 +8,5 @@ export interface ITemplateServiceClient {
     getTemplateParameters(templateId: string): Promise<ExtendedPipelineTemplate>;
     getTemplateConfiguration(templateId: string, inputs: StringMap<string>): Promise<ExtendedPipelineTemplate>;
     getTemplateFile(templateId: string, fileName: string): Promise<{ id: string, content: string }[]>;
+    getResourceFilteredTemplates(language: string, deployTarget: string): Promise<TemplateInfo[]>;
 }
--- a/src/configure/clients/ITemplateServiceClient.ts
+++ b/src/configure/clients/ITemplateServiceClient.ts
@@ -8,5 +8,5 @@ export interface ITemplateServiceClient {
     getTemplateParameters(templateId: string): Promise<ExtendedPipelineTemplate>;
     getTemplateConfiguration(templateId: string, inputs: StringMap<string>): Promise<ExtendedPipelineTemplate>;
     getTemplateFile(templateId: string, fileName: string): Promise<{ id: string, content: string }[]>;
-    getTemplatesInfoByFilter(language: string, deployTargetFilter?: string, buildTargetFilter?: string): Promise<TemplateInfo[]>;
+    getTemplatesInfoByFilter(resourceType: string): Promise<TemplateInfo[]>;
 }
--- a/src/configure/clients/github/TemplateServiceClient.ts
+++ b/src/configure/clients/github/TemplateServiceClient.ts
@@ -21,6 +21,22 @@ export class TemplateServiceClient implements ITemplateServiceClient {
         this.headers = headers;
     }
 
+    public async getResourceFilteredTemplates(language: string, deployTarget: string): Promise<TemplateInfo[]> {
+        return this.restClient.sendRequest(
+            {
+                url: this.templateServiceUri + this.templatesInfoResource,
+                method: 'GET',
+                headers: this.headers,
+                queryParameters: {
+                    'api-version': this.apiVersion,
+                    'languageFilter': language,
+                    'deployTargetFilter': deployTarget
+                },
+                deserializationMapper: null,
+                serializationMapper: null
+            });
+    }
+
     public async getTemplates(body: RepositoryAnalysis): Promise<TemplateInfo[]> {
         return this.restClient.sendRequest(
             {
--- a/src/configure/clients/github/TemplateServiceClient.ts
+++ b/src/configure/clients/github/TemplateServiceClient.ts
@@ -1,7 +1,7 @@
 import { RepositoryAnalysis } from "azureintegration-repoanalysis-client-internal";
 import { ServiceClientCredentials } from "ms-rest";
 import { ExtendedPipelineTemplate } from "../../model/Contracts";
-import { StringMap, TargetResourceType } from "../../model/models";
+import { StringMap } from "../../model/models";
 import { TemplateInfo } from "../../model/templateModels";
 import { ITemplateServiceClient } from "../ITemplateServiceClient";
 import { RestClient } from "../restClient";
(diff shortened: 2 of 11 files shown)


src/configure/helper/templateHelper.ts
```diff
@@ -151,6 +151,38 @@ export async function analyzeRepoAndListAppropriatePipeline(repoPath: string, re
...
+        remoteTemplates.forEach((templateInfo: TemplateInfo) => {
+            const remoteTemplate: RemotePipelineTemplate = {
+                label: templateInfo.templateLabel,
+                targetType: getTargetType(templateInfo),
+                targetKind: getTargetKind(templateInfo),
+                templateType: TemplateType.REMOTE,
+                language: templateInfo.attributes.language,
+                id: templateInfo.templateId,
+                templateWeight: templateInfo.templateWeight,
+                workingDirectory: templateInfo.workingDirectory,
```
anuragc617, line 172: this code is repeated in below function as well, can you create separate function for it with name convertToRemotePipelineTemplate

src/configure/clients/ITemplateServiceClient.ts
```diff
@@ -8,4 +8,5 @@ export interface ITemplateServiceClient {
     getTemplateParameters(templateId: string): Promise<ExtendedPipelineTemplate>;
     getTemplateConfiguration(templateId: string, inputs: StringMap<string>): Promise<ExtendedPipelineTemplate>;
     getTemplateFile(templateId: string, fileName: string): Promise<{ id: string, content: string }[]>;
+    getResourceFilteredTemplates(language: string, deployTarget: string): Promise<TemplateInfo[]>;
```
anuragc617, line 11: function name should be GetTemplatesInfoByFilter and provide all three args like this
getTemplatesInfoByFilter( languageFilter: string, deployTargetFilter?: string, buildTargetFilter?: string) : Promise<TemplateInfo[]>

src/configure/configure.ts
```diff
@@ -173,7 +173,11 @@ class Orchestrator {
                 this.inputs.azureSession = getAzureSession();
             }
             const repoAnalysisResult = await this.getRepositoryAnalysis();
-            await this.getSelectedPipeline(repoAnalysisResult);
+            if (resourceNode && resourceNode.type === TargetResourceType.AKS) {
```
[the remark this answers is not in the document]
kanika1894: Isn't it the same thing?
Anyway we'll have to put up a condition on the resourceNode type itself to put the required values for deployTarget and language, so instead of ``` context['isResourceSelected] && resourceNode.type === TargetResourceType.AKS ```, the existing one looke better?

src/configure/configure.ts
```diff
@@ -173,7 +173,11 @@ class Orchestrator {
                 this.inputs.azureSession = getAzureSession();
             }
             const repoAnalysisResult = await this.getRepositoryAnalysis();
-            await this.getSelectedPipeline(repoAnalysisResult);
+            if (this.context['isResourceAlreadySelected'] && resourceNode.type === TargetResourceType.AKS) {
+                await this.getFilteredTemplates("docker", "Azure:AKS");
```
bishal-pdMSFT, line 177: This will break the azure pipelines flow. `getSelectedPipeline` handles both local templates (which I thin is used for azure pipelines case) and remote templates (which is used for gh workflow case). It decides the kind of template to use based on the repo provider and some extension settings.
Hence it would not be wise to bifurcate here. We need to go deeper inside getSelectedPipeline and then only call getFilteredTemplates in case of remote templates.

kanika1894: ohh! yes! 
will make the required changes.

src/configure/clients/github/TemplateServiceClient.ts
```diff
@@ -21,6 +21,36 @@ export class TemplateServiceClient implements ITemplateServiceClient {
         this.headers = headers;
     }
 
+    public async getTemplatesInfoByFilter(resourceType: string): Promise<TemplateInfo[]> {
```
bishal-pdMSFT, line 24: this function should take language, build Target and deployTarget as parameters. The switch case needs to move to the caller

bishal-pdMSFT: The reason is that the client class should be as thin as possible and very specific to constructs of the service it is trying to  call. VS Code business constructs should not be much visible here.

src/configure/configure.ts
```diff
@@ -173,8 +182,12 @@ class Orchestrator {
                 this.inputs.azureSession = getAzureSession();
             }
             const repoAnalysisResult = await this.getRepositoryAnalysis();
-            await this.getSelectedPipeline(repoAnalysisResult);
-
+            await this.setPipelineType();
+            // Right click on not resource not supported for Azure pipeline
+            if (resourceNode && this.pipelineType === PipelineType.AzurePipeline) {
```
bishal-pdMSFT, line 187: it seems `resourceNode` is always going to be null as `analyzeNode()` always return null

src/configure/configure.ts
```diff
@@ -173,8 +182,12 @@ class Orchestrator {
...
+                throw Error("Scenario not supported for Azure pipelines.");
```
bishal-pdMSFT, line 188: Do we need to explicitly throw this error? Anyways getTemplatesByRepoAnalysis will return empty array of templates and hence a n error will be shown to user

src/configure/configure.ts
```diff
@@ -484,34 +498,22 @@ class Orchestrator {
         telemetryHelper.setTelemetry(TelemetryKeys.SubscriptionId, this.inputs.subscriptionId);
     }
 
-    private async getSelectedPipeline(repoAnalysisResult: RepositoryAnalysis): Promise<void> {
-        extensionVariables.templateServiceEnabled = true;
+    private async getTemplatesByRepoAnalysis(repoAnalysisResult: RepositoryAnalysis): Promise<void> {
+
         let appropriatePipelines: PipelineTemplate[] = [];
+        const remotePipelines: PipelineTemplate[] = await vscode.window.withProgress(
```
bishal-pdMSFT, line 504: I guess you have removed the check for `templateServiceEnabled`. But if you want to do so then you should do across the complete code and remove the variable definition as well. Otherwise it is not complete cleanup

src/configure/clients/github/TemplateServiceClient.ts
kanika1894: cool, will update

src/configure/configure.ts
kanika1894: line 268
```return this.inputs.targetResource.resource;```
It returns the resource in case of right click scenario.

File: src/configure/configure.ts (hunk 7 above)
Comment (kanika1894) [reply src/configure/configure.ts:188]: If the repo has something besides AKS, it will cause an issue and the flow will try to continue with webApp template and will break. So, anyway, we know that this scenario is not supported by us. therefore, explictly throwing. Better to have a proper error message to use instead of some dev exception.

File: src/configure/configure.ts (hunk 6 above)
Comment (bishal-pdMSFT) [reply src/configure/configure.ts:187]: Ah I missed that addition

File: src/configure/configure.ts [hunk 9]
```diff
@@ -252,6 +265,7 @@ class Orchestrator {
             if (await this.extractAzureResourceFromNode(node)) {
                 this.context['isResourceAlreadySelected'] = true;
                 this.context['resourceId'] = this.inputs.targetResource.resource.id;
+                return this.inputs.targetResource.resource;
```
bishal-pdMSFT, line 268: It looks weird to me that the function return a value in case it is a resource node but otherwise returns null if it is a folder node.
We should keep it generic and not return anything in either case. We are anyways setting `this.context['isResourceAlreadySelected'] = true;` which should be good way to check later whether resource is already selected.

src/configure/configure.ts
bishal-pdMSFT: As mentioned in another comment, instead of using `resourceNode` variable, we should rely on a function called `isResourceAlreadySelected` which will internally check if  this.context['isResourceAlreadySelected'] is true

kanika1894: yes, that we can do
Or anyway as we are initializing the targetResource as a part of inputs, that can be used as well.
Because for cases, where we require resource details when it is selected, we will have to use the ``` inputs.targetResource.resource ```

src/configure/configure.ts
kanika1894: updated

src/configure/configure.ts
```diff
@@ -159,24 +160,40 @@ class Orchestrator {
...
 
+    private async setPipelineType() {
+        if (this.inputs.sourceRepository.repositoryProvider === RepositoryProvider.Github && extensionVariables.enableGitHubWorkflow) {
+            this.pipelineType = PipelineType.GitHubPipeline;
+        } else {
+            this.pipelineType = PipelineType.AzurePipeline;
+        }
+    }
+
+    private async isResourceAlreadySelected(): Promise<boolean> {
```
bishal-pdMSFT, line 171: this should not be async

src/configure/configure.ts
bishal-pdMSFT: Did you handle this comment?

kanika1894: yes, it was there only in model.ts. Have removed it.

src/configure/configure.ts
```diff
@@ -159,24 +160,40 @@ class Orchestrator {
         }
     }
 
+    private async setPipelineType() {
```
bishal-pdMSFT, line 163: this should not be async

src/configure/configure.ts
bishal-pdMSFT: Okay, so after reading through and thinking a lot I think this is okay. I can justify this explicit condition so that we do not have to start filtering local templates by target resource type.

src/configure/configure.ts
```diff
@@ -159,24 +160,40 @@ class Orchestrator {
...
             if (!this.inputs.azureSession) {
                 this.inputs.azureSession = getAzureSession();
             }
             const repoAnalysisResult = await this.getRepositoryAnalysis();
-            await this.getSelectedPipeline(repoAnalysisResult);
-
+            await this.setPipelineType();
+            // Right click on not resource not supported for Azure pipeline
+            if (this.isResourceAlreadySelected() && this.pipelineType === PipelineType.AzurePipeline) {
+                throw Error("Scenario not supported for Azure pipelines.");
```
bishal-pdMSFT, line 192: I think you asked Pulkit for an error message. Is this final?

bishal-pdMSFT (approved):

src/configure/configure.ts
kanika1894: Still waiting for his approval. Will merge only after I get that.

bishal-pdMSFT: @kanika1894  Lets merge this PR so that you can do bugbash tomorrow. the string can anyways be changed easily later.

PR #173, opened 2020-10-03
Docker setting fix for right click scenario

tiwarishub:
Fix for issue https://github.com/github/azure-integration/issues/63

### DIFF
--- a/src/configure/configure.ts
+++ b/src/configure/configure.ts
@@ -556,17 +556,20 @@ class Orchestrator {
         if (!repoAnalysisResult || !repoAnalysisResult.applicationSettingsList) {
             return;
         }
-        const workingDirectories = Array.from(new Set(this.inputs.potentialTemplates
-            .filter((template) => template.templateType === TemplateType.REMOTE)
-            .map((template: RemotePipelineTemplate) => template.workingDirectory.toLowerCase())));
-        const applicationSettings = repoAnalysisResult.applicationSettingsList.filter((applicationSetting) => {
-            if (this.inputs.pipelineConfiguration.template.templateType === TemplateType.REMOTE) {
-                return workingDirectories.indexOf(applicationSetting.settings.workingDirectory.toLowerCase()) >= 0;
-            }
-            return applicationSetting.language === this.inputs.pipelineConfiguration.template.language;
 
-        });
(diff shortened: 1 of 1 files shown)


src/configure/configure.ts
```diff
@@ -556,17 +556,20 @@ class Orchestrator {
...
-            .filter((template) => template.templateType === TemplateType.REMOTE)
-            .map((template: RemotePipelineTemplate) => template.workingDirectory.toLowerCase())));
-        const applicationSettings = repoAnalysisResult.applicationSettingsList.filter((applicationSetting) => {
-            if (this.inputs.pipelineConfiguration.template.templateType === TemplateType.REMOTE) {
-                return workingDirectories.indexOf(applicationSetting.settings.workingDirectory.toLowerCase()) >= 0;
-            }
-            return applicationSetting.language === this.inputs.pipelineConfiguration.template.language;
 
-        });
+        let applicationSettings: ApplicationSettings[] = repoAnalysisResult.applicationSettingsList;
```
anuragc617, line 560: Is it working after this change?

anuragc617 (approved):

src/configure/configure.ts
tiwarishub: Yes, it worked.

src/configure/configure.ts
```diff
@@ -556,17 +556,20 @@ class Orchestrator {
...
+        if (!this.isResourceAlreadySelected()) {
+            const workingDirectories = Array.from(new Set(this.inputs.potentialTemplates
+                .filter((template) => template.templateType === TemplateType.REMOTE)
```
bishal-pdMSFT, line 563: @anuragc617 one question:  should we not be filtering more than just remote templates. We should only be interested in working directories of the templates which match the selected template attributes.

src/configure/configure.ts
```diff
@@ -556,17 +556,20 @@ class Orchestrator {
...
+                .map((template: RemotePipelineTemplate) => template.workingDirectory.toLowerCase())));
+            applicationSettings = repoAnalysisResult.applicationSettingsList.filter((applicationSetting) => {
+                if (this.inputs.pipelineConfiguration.template.templateType === TemplateType.REMOTE) {
```
bishal-pdMSFT, line 566: Could we have altered the fix to rather check for both Remote template as well valid working directory?

PR #175, opened 2020-10-05
Bugbash fixes with release changes

kanika1894:
Fixing https://github.com/github/azure-integration/issues/64
Fixing https://github.com/github/azure-integration/issues/62
Release changes for version 1.2.0

Also include changes for the following requirement :
When it's a azure or a local repo, we will give user the error, but in case of Github repo, even if the user has set azure pipeline in settings, we have to create a  github pipeline.

### DIFF
--- a/src/configure/configure.ts
+++ b/src/configure/configure.ts
@@ -22,7 +22,7 @@ import { Result, telemetryHelper } from './helper/telemetryHelper';
 import * as templateHelper from './helper/templateHelper';
 import { TemplateParameterHelper } from './helper/templateParameterHelper';
 import { ConfigurationStage } from './model/Contracts';
-import { extensionVariables, GitBranchDetails, GitRepositoryParameters, MustacheContext, ParsedAzureResourceId, QuickPickItemWithData, RepositoryProvider, SourceOptions, StringMap, TargetResourceType, WizardInputs } from './model/models';
+import { extensionVariables, GitBranchDetails, GitRepositoryParameters, MustacheContext, ParsedAzureResourceId, PipelineType, QuickPickItemWithData, RepositoryProvider, SourceOptions, StringMap, TargetResourceType, WizardInputs } from './model/models';
 import { DraftPipelineConfiguration, ProvisioningConfiguration, provisioningMode } from './model/provisioningConfiguration';
 import { LocalPipelineTemplate, PipelineTemplate, RemotePipelineTemplate, TemplateAssetType, TemplateType } from './model/templateModels';
 import * as constants from './resources/constants';
@@ -93,6 +93,7 @@ class Orchestrator {
     private controlProvider: ControlProvider;
     private continueOrchestration: boolean = true;
     private context: StringMap<any> = {};
+    private pipelineType: PipelineType;
 
     public constructor() {
         this.inputs = new WizardInputs();
@@ -159,13 +160,25 @@ class Orchestrator {
         }
     }
 
+    private setPipelineType() {
+        if (this.inputs.sourceRepository.repositoryProvider === RepositoryProvider.Github && extensionVariables.enableGitHubWorkflow) {
+            this.pipelineType = PipelineType.GitHubPipeline;
+        } else {
+            this.pipelineType = PipelineType.AzurePipeline;
+        }
+    }
+
+    private isResourceAlreadySelected(): boolean {
+        return this.context['isResourceAlreadySelected'];
+    }
+
     private async getInputs(node: any): Promise<void> {
         telemetryHelper.setTelemetry(TelemetryKeys.FF_UseGithubForCreatingNewRepository,
             vscode.workspace.getConfiguration().get('deployToAzure.UseGithubForCreatingNewRepository'));
         telemetryHelper.setTelemetry(TelemetryKeys.FF_UseAzurePipelinesForGithub,
             vscode.workspace.getConfiguration().get('deployToAzure.UseAzurePipelinesForGithub'));
 
-        const resourceNode = await this.analyzeNode(node);
+        await this.analyzeNode(node);
 
         if (this.continueOrchestration) {
             await this.getSourceRepositoryDetails();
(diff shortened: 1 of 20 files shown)


src/configure/helper/templateHelper.ts
```diff
@@ -176,45 +176,53 @@ export async function getFilteredTemplates(azureSession: AzureSession, resourceT
...
+                remoteTemplates = await getFilteredTemplates(azureSession, resource.type, githubPatToken);
+            }
+            else if (repoAnalysisParameters && repoAnalysisParameters.applicationSettingsList) {
+                remoteTemplates = await getTemplates(azureSession, repoAnalysisParameters, githubPatToken);
+            }
             pipelineTemplates = await convertToPipelineTemplate(remoteTemplates);
+            pipelineTemplates = pipelineTemplates.concat(localPipelineTemplates);
+            // sorted by weight
+            pipelineTemplates = pipelineTemplates.sort((a, b) => {
+                if (a.templateWeight < b.templateWeight) { return 1; }
```
anuragc617, line 216: return b.templateWeight - a.templateWeight. it will simplify the sort function

anuragc617 (approved):

src/configure/configure.ts
```diff
@@ -185,12 +186,26 @@ class Orchestrator {
...
+            }
+
             const repoAnalysisResult = await this.getRepositoryAnalysis();
-            this.setPipelineType();
-            // Right click on not resource not supported for Azure pipeline
-            if (this.isResourceAlreadySelected() && this.pipelineType === PipelineType.AzurePipeline) {
-                throw Error("Scenario not supported for Azure pipelines.");
+
+            // Right click on not resource not supported for local repo
+            if (this.isResourceAlreadySelected() && !repoAnalysisResult) {
```
bishal-pdMSFT, line 204: `!repoAnalysisResult` will also be false when it is a valid github remote but repo analysis was not able to detect anything (as it might not have support). We should not block such case and allow use to to proceed with a default Dockerfile path.

bishal-pdMSFT: May be we can set a flag to indicate local repo in `await this.getSourceRepositoryDetails();` and use that flag here?

src/configure/helper/templateHelper.ts
```diff
@@ -176,45 +178,52 @@ export async function getFilteredTemplates(azureSession: AzureSession, resourceT
     let filteredTemplates: TemplateInfo[];
     switch (resourceType) {
         case TargetResourceType.AKS:
-            filteredTemplates = await client.getTemplatesInfoByFilter("docker", "Azure:AKS");
+            await telemetryHelper.executeFunctionWithTimeTelemetry(async () => {
```
bishal-pdMSFT, line 181: we should use the defined function `getTemplates` here as well

src/configure/configure.ts
kanika1894: yeah, that can be done.

src/configure/helper/templateHelper.ts
kanika1894: getTemplates get templates based on repoAnlaysis result and this function is to get templates filtered on language, buildTarget and deployTarget.
Why should getTemplates be called here?

src/configure/configure.ts
```diff
@@ -185,12 +187,21 @@ class Orchestrator {
...
             }
+
+            // Right click scenario not supported for Azure and local repo
+            if (this.isResourceAlreadySelected()) {
+                if (this.inputs.sourceRepository.repositoryProvider === RepositoryProvider.AzureRepos || this.isLocalRepo) {
+                    throw Error(Messages.GithubRepoRequired);
+                } else if (!extensionVariables.enableGitHubWorkflow) {
+                    // For github repo, we create a github pipeline
+                    extensionVariables.enableGitHubWorkflow = true;
+                    vscode.window.showInformationMessage("Creating github pipeline.");
```
bishal-pdMSFT, line 198: This should be "Since this is right click scenario, the setting "<setting-name>" will be ignored"

src/configure/helper/templateHelper.ts
bishal-pdMSFT: yes,you are right

src/configure/helper/templateHelper.ts
```diff
@@ -176,45 +178,52 @@ export async function getFilteredTemplates(azureSession: AzureSession, resourceT
...
+                remoteTemplates = await getFilteredTemplates(azureSession, resource.type, githubPatToken);
+            }
+            else if (repoAnalysisParameters && repoAnalysisParameters.applicationSettingsList) {
+                remoteTemplates = await getTemplates(azureSession, repoAnalysisParameters, githubPatToken);
+            }
             pipelineTemplates = await convertToPipelineTemplate(remoteTemplates);
+            pipelineTemplates = pipelineTemplates.concat(localPipelineTemplates);
+            // sorted by weight
+            pipelineTemplates = pipelineTemplates.sort((a, b) => {
+                return b.templateWeight - a.templateWeight;
```
bishal-pdMSFT, line 218: you have changed the sorting algo. it was returning 1/-1 but now can return any integer. is this intentional and needed? @kanika1894

src/configure/configure.ts
```diff
@@ -379,6 +389,7 @@ class Orchestrator {
     }
 
     private setDefaultRepositoryDetails(): void {
+        this.isLocalRepo = true;
```
bishal-pdMSFT, line 392: This can be defined as extension variable so that can be used in other files/classes as well?

bishal-pdMSFT (approved):

src/configure/helper/templateHelper.ts
kanika1894: This is just to simplify the code.
It won't change the logic. The functioning of sort function is in such a way that it will just consider whether the difference is positive or negative, and sort accordingly.
More examples : https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort

src/configure/configure.ts
kanika1894: I don't think we are using this as of now, but yes, since can be used in future, will update.

src/configure/helper/templateHelper.ts
bishal-pdMSFT: Makes sense

PR #178, opened 2020-10-06
Bugbash 05/10 fixes

tiwarishub:
Bug fixes for 
https://github.com/github/azure-integration/issues/81
https://github.com/github/azure-integration/issues/72


src/configure/helper/LocalGitRepoHelper.ts
```diff
@@ -111,7 +111,6 @@ export class LocalGitRepoHelper {
      */
     public async addContentToFile(content: string, pathToFile: string): Promise<string> {
         fs.writeFileSync(pathToFile, content);
-        await vscode.workspace.saveAll(true);
```
tiwarishub, line 114: THis is done for all types of configurer. which i think is valid, we should not saved any unsaved files and writeFileSync as it is writing so it will save the file as well

src/configure/helper/templateHelper.ts
```diff
@@ -814,7 +814,7 @@ let azurePipelineTemplates: { [key in SupportedLanguage]: LocalPipelineTemplate[
 let githubWorklowTemplates: { [key in SupportedLanguage]: LocalPipelineTemplate[] } = {
     'docker': [
         {
-            label: 'Containerized application to AKS',
+            label: 'Containerized application to Azure Kubernetes Service (AKS)',
```
kanika1894, line 817: Hoping this label update is made in template service as well?
As we need to have the labels exactly same.

anuragc617: yes

src/configure/configurers/provisioningConfigurer.ts
```diff
@@ -163,7 +163,7 @@ export class ProvisioningConfigurer implements IProvisioningConfigurer {
     public async showPipelineFiles(): Promise<void> {
         this.filesToCommit.forEach(async (file) => {
             await this.localGitRepoHelper.addContentToFile(file.content, file.absPath);
-            await vscode.window.showTextDocument(vscode.Uri.file(file.absPath));
+            await vscode.window.showTextDocument(vscode.Uri.file(file.absPath), { preview: false });
```
anuragc617, line 166: what does the preview false do?

src/configure/helper/LocalGitRepoHelper.ts
anuragc617: @bishal-pdMSFT is this the required behaviour?

src/configure/helper/AssetHandler.ts
```diff
@@ -17,7 +17,7 @@ const Layer = "AssetCreationHandler";
 export class AssetHandler {
     // tslint:disable-next-line:no-reserved-keywords
     public async createAssets(assets: TemplateAsset[], inputs: WizardInputs, createAsset: (name: string, type: TemplateAssetType, data: any, inputs: WizardInputs) => Promise<string>): Promise<void> {
-        if (inputs.pipelineConfiguration.template.label === "Containerized application to AKS") {
+        if (inputs.pipelineConfiguration.template.label === "Containerized application to Azure Kubernetes Service (AKS)") {
```
kanika1894, line 20: Won't it be better if we specify this as a constant somewhere, so that reusing this string at multiple places becomes easier? This will also prevent mistype issues.

src/configure/helper/LocalGitRepoHelper.ts
bishal-pdMSFT: IMO in remote configurer, from a functionality point of view, we do not need files to be saved locally as we just need the content. There is no use of locally saved file. Infact that is a problem as we might to remove those files once remote provisioning has completed otherwise user will see those files as checked out.

bishal-pdMSFT: And I think the bug is also whether we should save just the workflow/manifest files rather than all files in workspace

tiwarishub: @bishal-pdMSFT , For this https://github.com/github/azure-integration/issues/72 bug we have removed the code to save  any unsaved files.
For the scenario where `git status` will show workflow files in untracked status, the only solution is i think we delete those workflow files from user workspace. The other approach like showing the files to users by showing only buffer with some vscode api, for this as well we need to create directory (.github or manifest, if this is first time) and adding adding empty files. In that case as well `git status` will show them as untracked files, even though they will be empty.

But other thing is , if we delete the workflow files then this is will be inconsistent with how our extension handles workflow files before remote configurer change,

bishal-pdMSFT: @tiwarishub lets merge without changing the existing behavior. This behavior is already shipped and hence not a blocker for the PR. This will unblock us to ship node changes.
We can take up this discussion separately and based on the conclusion create a separate PR.
cc @anuragc617

src/configure/configurers/provisioningConfigurer.ts
bishal-pdMSFT: I also want to know what does it do

bishal-pdMSFT (approved):

src/configure/helper/LocalGitRepoHelper.ts
tiwarishub: sure @bishal-pdMSFT

src/configure/configurers/provisioningConfigurer.ts
tiwarishub: When we set preview to false, it does not replace the content of current editor window. For example, if we don’t set preview to false and call function showTextDocument in a loop for different then same editor window will be used to show the files (not different editor window for different files)

PR #180, opened 2020-10-07
Right click support - changes with respect to AKS extension

kanika1894:
Updating node property keys as per the node structure defined in aks extension

### DIFF
--- a/src/configure/configure.ts
+++ b/src/configure/configure.ts
@@ -275,16 +275,14 @@ class Orchestrator {
 
     private async analyzeNode(node: any): Promise<void> {
         if (!!node) {
-            if (await this.extractAzureResourceFromNode(node)) {
+            if (node.fsPath) {
+                //right click on a folder
+                this.workspacePath = node.fsPath;
+                telemetryHelper.setTelemetry(TelemetryKeys.SourceRepoLocation, SourceOptions.CurrentWorkspace);
+            } else if (await this.extractAzureResourceFromNode(node)) {
                 // right click on a resource
                 this.context['isResourceAlreadySelected'] = true;
                 this.context['resourceId'] = this.inputs.targetResource.resource.id;
-            } else {
-                if (node.fsPath) {
-                    //right click on a folder
-                    this.workspacePath = node.fsPath;
-                    telemetryHelper.setTelemetry(TelemetryKeys.SourceRepoLocation, SourceOptions.CurrentWorkspace);
-                }
             }
         }
     }
@@ -456,7 +454,7 @@ class Orchestrator {
     private async extractAzureResourceFromNode(node: IResourceNode): Promise<boolean> {
         if (!!node.resource.id && node.resource.type != 'cluster') {
             this.inputs.subscriptionId = node.subscriptionId;
-            this.inputs.azureSession = getSubscriptionSession(this.inputs.subscriptionId);
+            this.inputs.azureSession = await getSubscriptionSession(this.inputs.subscriptionId);
             this.azureResourceClient = new AppServiceClient(this.inputs.azureSession.credentials, this.inputs.azureSession.environment, this.inputs.azureSession.tenantId, this.inputs.subscriptionId);
 
             try {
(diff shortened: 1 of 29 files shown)


src/configure/configure.ts
```diff
@@ -478,12 +478,12 @@ class Orchestrator {
                 throw error;
             }
         }
-        else if (!!node.cloudResource && node.cloudResource.nodeType === 'cluster') {
-            this.inputs.subscriptionId = node.cloudResource.subscription.subscriptionId;
+        else if (node.resource.nodeType === "cluster") {
```
bishal-pdMSFT, line 481: We should use same interface which we created in AKS extension. The node object should be of that type instead of AzureTreeitem

bishal-pdMSFT: In fact we should use this type when we are registering the command itself in https://github.com/microsoft/vscode-deploy-azure/blob/fdf6af7893f9eadc59c79e0bd1636bc01931b9be/src/configure/activate.ts#L25

bishal-pdMSFT: We will need to use `IResourceNode | any` to accomodate folder right click.

Can we also model right click as an interface instead of `any`?

bishal-pdMSFT, line 481: This function should bifurcate based on whether it is a resource right click or folder right click and that should happen based on the type of `node` first.

kanika1894: The bifurcation is at ``` analyzeNode ``` function. 
It eneters this section only if it's a right click on scenario.

kanika1894: The folder structure is way different from the resource structure. So at the initial stage, have kept ``` any ``` and where we are dealing the the resource scenario, have used ``` IResourceNode ```.

src/configure/configure.ts
```diff
@@ -280,15 +280,14 @@ class Orchestrator {
 
     private async analyzeNode(node: any): Promise<void> {
```
bishal-pdMSFT, line 281: Can we have a type for folder node as well? VS Code must have defined a type for that. We should use that if possible.

If so, then we can defined the argument type here as `IResourceNode | FolderNode`. (replace`FolderNode` with the type for folder)

src/configure/configure.ts
```diff
@@ -280,15 +280,14 @@ class Orchestrator {
...
         if (!!node) {
-            if (await this.extractAzureResourceFromNode(node)) {
+            if (node.fsPath) {
```
bishal-pdMSFT, line 283: If we do above, then we can check the type here using `typeof`. This makes the code easier to understand

src/configure/configure.ts
```diff
@@ -37,7 +37,7 @@ const uuid = require('uuid/v4');
 const Layer: string = 'configure';
 export let UniqueResourceNameSuffix: string = uuid().substr(0, 5);
 
-export async function configurePipeline(node: AzureTreeItem) {
+export async function configurePipeline(node: any) {
```
bishal-pdMSFT, line 40: use same type here as well

src/configure/configure.ts
```diff
@@ -278,17 +278,17 @@ class Orchestrator {
         return this.inputs.potentialTemplates[0].targetType;
     }
 
-    private async analyzeNode(node: any): Promise<void> {
+    private async analyzeNode(node: IResourceNode | vscode.Uri): Promise<void> {
         if (!!node) {
-            if (await this.extractAzureResourceFromNode(node)) {
+            const folderNode = node as vscode.Uri;
+            if (folderNode && folderNode.fsPath) {
```
bishal-pdMSFT, line 284: ohh, it worked! what is the different compared to what you tried during call?

src/configure/configure.ts
```diff
@@ -278,17 +278,17 @@ class Orchestrator {
...
+                // right click on a folder
+                this.workspacePath = folderNode.fsPath;
+                telemetryHelper.setTelemetry(TelemetryKeys.SourceRepoLocation, SourceOptions.CurrentWorkspace);
+            } else if (await this.extractAzureResourceFromNode(node as IResourceNode)) {
```
bishal-pdMSFT, line 288: And here the `as` works without assigning first to a local variable?

bishal-pdMSFT (approved):

src/configure/configure.ts
kanika1894: ``` node as vscode.Uri ``` will still be required and then if() will just check fspath.

src/configure/configure.ts
kanika1894: yeah, it is just passing the variable, so no issue.

PR #182, opened 2020-10-07
Ensuring we replace using the index.

ashwinsangem:


src/configure/utilities/templateConverter.ts
```diff
@@ -22,9 +22,10 @@ export function convertExpression(expression: string): string {
         helperName = expression.substring(openingHelperFunc.length + 1, expression.indexOf(' '));
         parts.push(expression.substring(0, expression.indexOf(' ')) + closingHelperFunc);
         let part = expression.substring(expression.indexOf(' '), expression.indexOf(closingHelperFunc)).trim();
-        let inputs = part.match(parametersRegex) || [];
-        for (let input of inputs) {
-            part = part.replace(input, "{{{" + input + "}}}")
+        let input = parametersRegex.exec(part);
```
bishal-pdMSFT, line 25: @anuragc617 can you please review this part

anuragc617: @bishal-pdMSFT  yes this looks good. But the template service change will be a breaking change as it will not work for the already released deploy to azure extension because of this bug.

anuragc617 (approved):

src/configure/utilities/templateConverter.ts
ashwinsangem: Yes, this code needs to go first to prod.

bishal-pdMSFT: Yes, thats true. We need to be careful here. @ashwinsangem lets sync up how we can deal with this. I guess we need versioning sooner than later. But lets plan what we can do in short term.

bishal-pdMSFT: BTW @ashwinsangem I believe this PR should go in regardless of other things.

PR #184, opened 2020-10-08
Caching the datasource get calls for a CD set up.

anuragc617:


src/configure/templateInputHelper/utilities/DataSourceUtility.ts
```diff
@@ -38,17 +39,23 @@ export class DataSourceUtility {
...
-        var requestBody = !!dataSource.requestBody ? MustacheHelper.render(dataSource.requestBody, view) : null;
-        let amrClient = new ArmRestClient(azureSession);
-        return amrClient.fetchArmData(armUri, httpMethod, JSON.parse(requestBody))
-            .then((response: any) => {
-                return this.evaluateDataSourceResponse(dataSource, response, view);
-            });
+        const requestBody = !!dataSource.requestBody ? MustacheHelper.render(dataSource.requestBody, view) : null;
+        const armClient = new ArmRestClient(azureSession);
+        let result: any;
+        if (httpMethod == "GET" && Cache.getCache().getValue(armUri)) {
```
vineetmimrot, line 45: Instead of checking at everyplace, whether the request is cached or not, Should we create a cache client which will check the URL in the map and if it is already processed then return cache result like https://www.npmjs.com/package/cacheable-request

PR #186, opened 2020-10-09
Fixing multi repo url for local configurer

anuragc617:


kanika1894 (commented):
Can you please provide more details with the PR? What scenario are we trying to handle.

src/configure/configurers/localGithubWorkflowConfigurer.ts
```diff
@@ -258,8 +258,6 @@ export class LocalGitHubWorkflowConfigurer implements Configurer {
                         metadata["properties"]["configName"] = `${doc["name"]}`;
                     }
                     metadata["properties"]["configPath"] = `${configPath}`;
-                    metadata["properties"]["repoUrl"] = `https://github.com/${inputs.sourceRepository.repositoryId}`;
```
bishal-pdMSFT, line 261: This is in case of local configurer? If so then where will these be set?

anuragc617: It is being set as part of source control api. This was added to make it consistent to provisioning service but this results in a double repo url because of some bug in source control api. In provisioning service we are updating first metadata and scmtype and then the sourcecontrol so we are not facing this issue for deployment configured using provisioning service. But here we are doing, first we are doing source control update and then metadata update which is causing this issue.

vineetmimrot: Should we be consistent here in both local and remote configurer. we can use the same order which provisioning svc is using.

anuragc617: After changing the order, it works fine. The error is happening because of repoUrl key case mismatch for different web apps.

src/configure/configurers/localGithubWorkflowConfigurer.ts
```diff
@@ -273,12 +272,9 @@ export class LocalGitHubWorkflowConfigurer implements Configurer {
 
                 let authorName = await LocalGitRepoHelper.GetHelperInstance(inputs.sourceRepository.localPath).getUsername();
                 let deployerName = 'GITHUBACTION';
-                let updateDeploymentLogPromise = (azureResourceClient as AppServiceClient).publishDeploymentToAppService(inputs.targetResource.resource.id, deploymentMessage, authorName, deployerName);
+                await (azureResourceClient as AppServiceClient).publishDeploymentToAppService(inputs.targetResource.resource.id, deploymentMessage, authorName, deployerName);
```
bishal-pdMSFT, line 275: Can we do sourcecontrol and deployment update in parallel?

src/configure/configurers/localGithubWorkflowConfigurer.ts
```diff
@@ -273,12 +272,9 @@ export class LocalGitHubWorkflowConfigurer implements Configurer {
...
 
-                Q.all([updateMetadataPromise, updateDeploymentLogPromise])
-                    .then(() => {
-                        telemetryHelper.setTelemetry(TelemetryKeys.UpdatedWebAppMetadata, 'true');
-                    });
+                telemetryHelper.setTelemetry(TelemetryKeys.UpdatedWebAppMetadata, 'true');
```
bishal-pdMSFT, line 277: the telemtries need to get fixed

bishal-pdMSFT: add for sourcecontrol as well

bishal-pdMSFT: Or you think just one if enough ?

src/configure/configurers/localGithubWorkflowConfigurer.ts
anuragc617: All the post steps are done in background. So I thought to serialize it to keep things simple. I have updated it to make the request in parallel

src/configure/configurers/localGithubWorkflowConfigurer.ts
anuragc617: Added source control telemetry as well

src/configure/configurers/localGithubWorkflowConfigurer.ts
```diff
@@ -274,10 +274,10 @@ export class LocalGitHubWorkflowConfigurer implements Configurer {
                 let authorName = await LocalGitRepoHelper.GetHelperInstance(inputs.sourceRepository.localPath).getUsername();
                 let deployerName = 'GITHUBACTION';
                 let updateDeploymentLogPromise = (azureResourceClient as AppServiceClient).publishDeploymentToAppService(inputs.targetResource.resource.id, deploymentMessage, authorName, deployerName);
-
-                Q.all([updateMetadataPromise, updateDeploymentLogPromise])
+                Q.all([sourceControlPromise, updateDeploymentLogPromise])
                     .then(() => {
                         telemetryHelper.setTelemetry(TelemetryKeys.UpdatedWebAppMetadata, 'true');
+                        telemetryHelper.setTelemetry(TelemetryKeys.UpdateWepAppSourceControl, 'true');
```
bishal-pdMSFT, line 280: I think we should just have one telemetry to log. Does not make sense to log two things for same point in code execution.
May be it was okay the way it was.

bishal-pdMSFT (approved):

PR #191, opened 2020-10-20
Workflow in tmp dir

tiwarishub:
This change is done to show draft workflow files in Temp or tmp directory instead of  adding them to local repo . Below is screenshot how workflow will be shown.
image

Once user clicks on "Modify and Commit" button, we will remove the workflow files from tmp directory  and add `(deleted)` for the active editor tabs, for example like this 
image
In the above `*-1365.yml` files are deleted one.

# Note :
   A good user interface for user will be , we should close/hide the text editors for tmp files  as those files are deleted but vscode does not expose any api for for open text editor which can be used to close a particular text editor. They only thing they provide is  `commands.executeCommand(workbench.action.closeActiveEditor)`  or `commands.executeCommand('workbench.action.closeAllEditors')` , which will not fulfil our scenarios.
They have long pending feature request (almost 3 years), but it is still not delivered. https://github.com/microsoft/vscode/issues/21617
https://github.com/microsoft/vscode/issues/15178

If user discards the workflow, then we will move the workflow files from tmp  directory to  git repo directory. then will again will open those workflows,

image
Please note from above pic is, now each created workflow file is there in two tabs of vscode, One with `(deleted)` and other one is which is moved.

image

Last , if user want to browse the committed workflow files, then we have added one information messages for that as well.
image


bishal-pdMSFT:
Cool description @tiwarishub ❤️

bishal-pdMSFT:
Few suggestion based on description:
1. Should we add `.github` folder in temp as well and keep same folder layout? This way it might easier for user to co-relate the structure of the files.
2. Suggestion to try: Does vs code allow closing a editor by its handle? We would get the handle when we open the editor. Just a thought.
3. The two notifications are looking pretty same and hence confusing. Can the workflow file related message be made part of same notification? We can add more text to convey this and may be make the file url as a hyperlink text.

tiwarishub:
> Few suggestion based on description:
> 
> 1. Should we add `.github` folder in temp as well and keep same folder layout? This way it might easier for user to co-relate the structure of the files.
> 2. Suggestion to try: Does vs code allow closing a editor by its handle? We would get the handle when we open the editor. Just a thought.
> 3. The two notifications are looking pretty same and hence confusing. Can the workflow file related message be made part of same notification? We can add more text to convey this and may be make the file url as a hyperlink text.

1. Yes @bishal-pdMSFT , now workflow files will be in .github and manifest will be in dockerContent in tmp directory
2. vscode only exposes `export let activeTextEditor: TextEditor | undefined;` , which is basically which is active editor. But in TextEditor type there is no way to close that editor (they had hide() function but now they have deprecated it because of its unusual behaviour ). 
3. Now i have removed 2 information box and made the commit link as hyperlink
image

bishal-pdMSFT:
For the notification message, I feel it is very obscure and not intuitive for user to understand that it will take to workflow file. We can add explicit text to tell user that a workflow file (with hypelink file name) has been pushed to their repo.
Take @pulkitaggarwl 's help to get the string.

src/configure/configure.ts
```diff
@@ -695,8 +695,8 @@ class Orchestrator {
             await provisioningConfigurer.postSteps(provisioningConfiguration, (draftProvisioningPipeline.result.pipelineConfiguration as DraftPipelineConfiguration), this.inputs);
 
             // All done, now browse the pipeline
-            telemetryHelper.setCurrentStep('BrowsingPipeline');
-            await provisioningConfigurer.browseQueuedPipeline();
+            telemetryHelper.setCurrentStep('BrowseQueuedWorkflow');
```
bishal-pdMSFT, line 698: This telemetry might already be in use in existing queries/dashboards. Please do not change unles absolute necessary.

src/configure/configurers/provisioningConfigurer.ts
```diff
@@ -185,15 +192,14 @@ export class ProvisioningConfigurer implements IProvisioningConfigurer {
...
+        let directoryPath: string = this.tmpDirectoryPath;
         dirList.forEach((dir) => {
             try {
                 directoryPath = Path.join(directoryPath, dir);
                 // tslint:disable-next-line:non-literal-fs-path
-                if (!fs.existsSync(directoryPath)) {
+                if (!fse.existsSync(directoryPath)) {
                     // tslint:disable-next-line:non-literal-fs-path
-                    fs.mkdirSync(directoryPath);
+                    fse.mkdirSync(directoryPath);
```
bishal-pdMSFT, line 202: Not related to your change, but is there no flavor of mkdir which creates all intermediate directories? Then we won't need to create all of them in a loop.

src/configure/configurers/provisioningConfigurer.ts
```diff
@@ -164,18 +169,20 @@ export class ProvisioningConfigurer implements IProvisioningConfigurer {
...
 
     public setQueuedPipelineUrl(provisioningConfiguration: ProvisioningConfiguration, inputs: WizardInputs) {
         const commitId = (provisioningConfiguration.result.pipelineConfiguration as CompletePipelineConfiguration).commitId;
         this.queuedPipelineUrl = `https://github.com/${inputs.sourceRepository.repositoryId}/commit/${commitId}/checks`;
+        this.committedWorkflow = `https://github.com/${inputs.sourceRepository.repositoryId}/commit/${commitId}`;
     }
 
     public async getFileToCommit(draftPipelineConfiguration: DraftPipelineConfiguration): Promise<void> {
         let destination: string;
+        this.tmpDirectoryPath = fse.mkdtempSync(os.tmpdir());
```
bishal-pdMSFT, line 185: This function seems to be also available in `fs` package. Any advantage of using `fse`?

bishal-pdMSFT, line 185: It seems we need to give a separator after `os.tmpdir()`. From docs:
```
The fs.mkdtemp() method will append the six randomly selected characters directly to the prefix string. For instance, given a directory /tmp, if the intention is to create a temporary directory within /tmp, the prefix must end with a trailing platform-specific path separator (require('path').sep).
```

src/configure/configurers/provisioningConfigurer.ts
```diff
@@ -36,6 +37,8 @@ export class ProvisioningConfigurer implements IProvisioningConfigurer {
     private maxNonStatusRetry: number = 60; // retries for max 5 min
     private localGitRepoHelper: LocalGitRepoHelper;
     private filesToCommit: DraftFile[] = [];
+    private committedWorkflow: string;
+    private tmpDirectoryPath: string;
```
bishal-pdMSFT, line 41: give a better name of what this temp is for. for example `tempWorkflowDirPath`

src/configure/configurers/provisioningConfigurer.ts
```diff
@@ -164,18 +169,20 @@ export class ProvisioningConfigurer implements IProvisioningConfigurer {
...
         });
     }
 
     public setQueuedPipelineUrl(provisioningConfiguration: ProvisioningConfiguration, inputs: WizardInputs) {
         const commitId = (provisioningConfiguration.result.pipelineConfiguration as CompletePipelineConfiguration).commitId;
         this.queuedPipelineUrl = `https://github.com/${inputs.sourceRepository.repositoryId}/commit/${commitId}/checks`;
+        this.committedWorkflow = `https://github.com/${inputs.sourceRepository.repositoryId}/commit/${commitId}`;
     }
 
     public async getFileToCommit(draftPipelineConfiguration: DraftPipelineConfiguration): Promise<void> {
```
bishal-pdMSFT, line 183: please rename to `populateFilesToCommit` as we are not returning anything. we are populating an object.

src/configure/configurers/provisioningConfigurer.ts
```diff
@@ -185,15 +192,14 @@ export class ProvisioningConfigurer implements IProvisioningConfigurer {
 
     public async getPathToFile(fileName: string, directory: string) {
```
bishal-pdMSFT, line 193: make this function private

bishal-pdMSFT: And put all public functions at top and all private ones at bottom

src/configure/configurers/provisioningConfigurer.ts
```diff
@@ -288,4 +293,13 @@ export class ProvisioningConfigurer implements IProvisioningConfigurer {
             files,
         } as DraftPipelineConfiguration;
     }
+
+    private async moveWorkflowFilesToRepoLocally(): Promise<void> {
+        const directoryPath: string = await this.localGitRepoHelper.getGitRootDirectory();
```
bishal-pdMSFT, line 298: rename to `gitRootDirectory`. usually a very common name like `directoryPath` does not convey precise info to reader

src/configure/configurers/provisioningConfigurer.ts
```diff
@@ -288,4 +293,13 @@ export class ProvisioningConfigurer implements IProvisioningConfigurer {
             files,
         } as DraftPipelineConfiguration;
     }
+
+    private async moveWorkflowFilesToRepoLocally(): Promise<void> {
```
bishal-pdMSFT, line 297: `moveWorkflowFilesToLocalRepo`

src/configure/configurers/provisioningConfigurer.ts
```diff
@@ -288,4 +293,13 @@ export class ProvisioningConfigurer implements IProvisioningConfigurer {
...
+        const directoryPath: string = await this.localGitRepoHelper.getGitRootDirectory();
+        for (const file of this.filesToCommit) {
+            fse.moveSync(file.absPath, Path.join(directoryPath, file.path));
+            await vscode.window.showTextDocument(vscode.Uri.file(Path.join(directoryPath, file.path)), { preview: false });
```
bishal-pdMSFT, line 301: `Path.join(directoryPath, file.path)` is re-used. Please move to a local variable first.

src/configure/configurers/provisioningConfigurer.ts
```diff
@@ -131,9 +134,11 @@ export class ProvisioningConfigurer implements IProvisioningConfigurer {
                 });
         } else {
             telemetryHelper.setTelemetry(TelemetryKeys.PipelineDiscarded, 'true');
+            await this.moveWorkflowFilesToRepoLocally();
             throw new UserCancelledError(Messages.operationCancelled);
         }
 
+        fse.removeSync(this.tmpDirectoryPath);
```
bishal-pdMSFT, line 141: Will this also remove the files present in temp dir?

src/configure/configurers/provisioningConfigurer.ts
```diff
@@ -185,15 +192,14 @@ export class ProvisioningConfigurer implements IProvisioningConfigurer {
 
     public async getPathToFile(fileName: string, directory: string) {
         const dirList = directory.split("/"); // Hardcoded as provisioning service is running on linux and we cannot use Path.sep as it is machine dependent
-        let directoryPath: string = "";
-        directoryPath = await this.localGitRepoHelper.getGitRootDirectory();
+        let directoryPath: string = this.tmpDirectoryPath;
         dirList.forEach((dir) => {
             try {
                 directoryPath = Path.join(directoryPath, dir);
                 // tslint:disable-next-line:non-literal-fs-path
```
bishal-pdMSFT, line 199: this rule does not seem to be practical. we should disable this rule in lint configuration.

src/configure/helper/LocalGitRepoHelper.ts
```diff
@@ -60,6 +47,23 @@ export class LocalGitRepoHelper {
...
+
+    public async getUsername(): Promise<string> {
+        let username = await this.gitReference.raw([
+            'config',
+            'user.name'
+        ]);
+
+        return username;
+    }
+
```
bishal-pdMSFT, line 66: Any specific reason for moving functions around.
Keep field declaration at top of the class
Keep constructor at the top of the class

src/configure/helper/LocalGitRepoHelper.ts
```diff
@@ -202,4 +198,13 @@ export class LocalGitRepoHelper {
         const buf = fs.readFileSync(pathToFile);
         return buf.toString();
     }
+
+    public async addContentToFile2(content: string, pathToFile: string): Promise<string> {
+        fs.writeFileSync(pathToFile, content);
```
bishal-pdMSFT, line 203: this is a single statement and can be made inline. Does not require a separate function.
If at all you want a separate function, then give a better name.

src/configure/configure.ts
tiwarishub: Done

src/configure/configurers/provisioningConfigurer.ts
tiwarishub: Yes we do have ``mkdirpSync`` in fs-extra module, but we need keep this loop because path is coming from provisioning service (which is on unix) and we need to create path based on user os so we cannot pass the same path to ``mkdirpSync`` function

src/configure/configurers/provisioningConfigurer.ts
tiwarishub: `fs-extra` basically extends the functionally ``fs`` provides. It contains all ``fs`` module function plus some additional function

tiwarishub: done

src/configure/configurers/provisioningConfigurer.ts
tiwarishub: done

src/configure/configurers/provisioningConfigurer.ts
tiwarishub: done

src/configure/configurers/provisioningConfigurer.ts
tiwarishub: done

src/configure/configurers/provisioningConfigurer.ts
tiwarishub: done

src/configure/configurers/provisioningConfigurer.ts
tiwarishub: done

src/configure/configurers/provisioningConfigurer.ts
tiwarishub: done

src/configure/configurers/provisioningConfigurer.ts
tiwarishub: Yes

src/configure/configurers/provisioningConfigurer.ts
tiwarishub: added in tslint

src/configure/helper/LocalGitRepoHelper.ts
tiwarishub: This file had  lot of linting error. And i was adding one function in it so though to clean this file as well.
Static functions should come before any non-static field and pub/private function.

src/configure/helper/LocalGitRepoHelper.ts
tiwarishub: Actually ``LocalGitRepoHelper`` class we use to handle things related to files and dir operations. So i have made this function to keeps consistency. I have rename this function to ``writeFileContent``

tiwarishub:
Hi @pulkitaggarwl ,
  Need your help in finalizing the information box to be shown to user for tmp directory changes , which we had discussed with you. Below are some of options , Please let us know if  any one of them is fine or if have any suggestion for different one.  Commit files url we have made hyperlink (in 2 and 3rd screenshot) , so that user can browse commited files as well

1. 
image

2.  
image

3. 
image

pulkitaggarwl:
image

This format. change "commit url link" to  "commit url" and hyperlink  "commit url"

tiwarishub:
Changed commit URL
image

src/configure/configurers/provisioningConfigurer.ts
```diff
@@ -288,4 +279,25 @@ export class ProvisioningConfigurer implements IProvisioningConfigurer {
...
+        }
+        fse.removeSync(this.tempWorkflowDirPath);
+    }
+
+    private async getPathToFile(fileName: string, directory: string) {
+        const dirList = directory.split("/"); // Hardcoded as provisioning service is running on linux and we cannot use Path.sep as it is machine dependent
+        let directoryPath: string = this.tempWorkflowDirPath;
+        dirList.forEach((dir) => {
+                directoryPath = Path.join(directoryPath, dir);
+         });
```
bishal-pdMSFT, line 298: As easier way could it to call `join`

```
normalizedPath = dirList.join(path.sep);
directoryPath = Path.join(directoryPath, normalizedPath);


```

bishal-pdMSFT (approved):

PR #193, opened 2020-11-10
VS code integration with local Proxy server

tiwarishub:


src/configure/helper/remoteServiceUrlHelper.ts
```diff
@@ -18,6 +18,14 @@ export class RemoteServiceUrlHelper {
     public static provisioningServiceRedirectUrl: string = "https://go.microsoft.com/fwlink/?linkid=2142042";
 
     public static async getTemplateServiceDefinition(): Promise<IServiceUrlDefinition> {
+        const deployment = process.env["DEPLOY_TO_AZURE_DEPLOYMENT_ENV"];
```
anuragc617, line 21: env name should DEPLOY_TO_AZURE_DEV_ENV

anuragc617: Got it. 
Rename deployment to environment.

anuragc617 (approved):

src/configure/helper/remoteServiceUrlHelper.ts
```diff
@@ -18,6 +18,14 @@ export class RemoteServiceUrlHelper {
     public static provisioningServiceRedirectUrl: string = "https://go.microsoft.com/fwlink/?linkid=2142042";
 
     public static async getTemplateServiceDefinition(): Promise<IServiceUrlDefinition> {
+        const deployment = process.env["DEPLOY_TO_AZURE_ENVIRONMENT"];
+        if (deployment != undefined && deployment === "development") {
+            return {
+                serviceFramework: ServiceFramework.Moda,
+                serviceUrl: "http://localhost:12345/"
```
bishal-pdMSFT, line 25: As suggested in the other PR, this url can also be made an env variable

src/configure/helper/remoteServiceUrlHelper.ts
```diff
@@ -18,6 +18,14 @@ export class RemoteServiceUrlHelper {
     public static provisioningServiceRedirectUrl: string = "https://go.microsoft.com/fwlink/?linkid=2142042";
 
     public static async getTemplateServiceDefinition(): Promise<IServiceUrlDefinition> {
+        const deployment = process.env["DEPLOY_TO_AZURE_ENVIRONMENT"];
```
bishal-pdMSFT, line 21: A more precise name would be `DEPLOY_ENVIRONMENT` as Azure is just another environment

src/configure/helper/remoteServiceUrlHelper.ts
```diff
@@ -18,6 +18,14 @@ export class RemoteServiceUrlHelper {
...
+        if (deployment != undefined && deployment === "development") {
```
bishal-pdMSFT, line 22: Is there a way to make this check more stricter? I mean is there a way to let this happen only when extension is running locally and not via vsix.
I am thinking from security point of view and it could open up a way for a bad user to make the flow go to a spoofed url and dupe a good user

bishal-pdMSFT (approved):

src/configure/helper/remoteServiceUrlHelper.ts
tiwarishub: There are chances where we might be taking extension from market place and setting env in our local and using to test to locally deployed service. Then we will not be able differentiate. To avoid we have env name DEPLOY_TO_AZURE_EXT_ENVIRONMENT as unique as possible

src/configure/helper/remoteServiceUrlHelper.ts
tiwarishub: I have changed it to DEPLOY_TO_AZURE_EXT_ENVIRONMENT  because DEPLOY_ENVIRONMENT is little common name so there may be change if any user is using it and it has this env set

PR #199, opened 2020-12-07
Fixing bug - Azure pipeline with GH repo

kanika1894:
In the scenario where we are creating Azure pipeline using a Github repo, there is an error in the flow where it tries to run the build. 
image
This was occuring as the .yaml file was getting checked in at the working directory but for the build, the Az pipeline was looking at the root location which was a mismatch, hence causing the error.

The requirement for the Azure pipeline flow is to have the .yaml file at the root folder and therefore, making the path change so that the pipeline is able to fetch the right file and the build can be run.


bishal-pdMSFT:
@kanika1894 is this PR not required?

kanika1894:
> @kanika1894 is this PR not required?

No, please refer this one :https://github.com/microsoft/vscode-deploy-azure/pull/200/

PR #200, opened 2020-12-10
Fixing bug - Azure pipeline with GH repo

kanika1894:
In the scenario where we are creating Azure pipeline using a Github repo, there is an error in the flow where it tries to run the build.
image

This was occuring as the .yaml file was getting checked in at the working directory but for the build, the Az pipeline was looking at the root location which was a mismatch, hence causing the error.

The requirement for the Azure pipeline flow is to have the .yaml file at the root folder and therefore, making the path change so that the pipeline is able to fetch the right file and the build can be run.


src/configure/configurers/azurePipelineConfigurer.ts
```diff
@@ -386,4 +387,4 @@ export class AzurePipelineConfigurer implements Configurer {
         let aadApp = await GraphHelper.createSpnAndAssignRole(inputs.azureSession, aadAppName, scope);
         return await serviceConnectionHelper.createAzureSPNServiceConnection(serviceConnectionName, inputs.azureSession.tenantId, inputs.subscriptionId, scope, aadApp);
     }
-}
+}
```
anuragc617, line 390: give a line after this

src/configure/helper/devOps/azureDevOpsHelper.ts
```diff
@@ -117,7 +117,7 @@ export class AzureDevOpsHelper {
             },
             process: {
                 type: 2,
-                yamlFileName: path.join(inputs.pipelineConfiguration.workingDirectory, path.basename(inputs.pipelineConfiguration.filePath))
+                yamlFileName: path.basename(inputs.pipelineConfiguration.filePath)
```
anuragc617, line 120: logic here and while detecting file path to commit is different because of which this bug has occurred. I think better option is inputs.pipelineConfiguration.filePath should contain the filePath in w.r.t. to root of repository. And join local rootDirectoryPath to inputs.pipelineConfiguration.filePath while adding file to localGit.

anuragc617 (approved):

src/configure/helper/devOps/azureDevOpsHelper.ts
kanika1894: As discussed, since the requirement is to have azure pipeline files always at the root level, keeping it this way and adding a comment in the code for better understanding.

src/configure/configurers/azurePipelineConfigurer.ts
kanika1894: done

PR #202, opened 2020-12-17
VScode is caching template and provisioning service  clients first time when deploy to azure command runs

tiwarishub:
VScode extension deactivates method is only called when vscode window is closed https://github.com/Microsoft/vscode/issues/47881 , And we are caching the client so for each different requests same PAT we are sending in the request.  So we added reset method which will be called whenever we start using extension.


src/configure/clients/TemplateServiceClientFactory.ts
```diff
@@ -24,5 +24,9 @@ export class TemplateServiceClientFactory {
         return this.client;
     }
 
+    public static resetClient() {
```
bishal-pdMSFT, line 27: This approach has one problem - it exposes the reset functionality to outside the factory. And that can be problematic if rest of the code uses reset in an irresponsible way which can result in client becoming useless midway.

bishal-pdMSFT: Al alternative, if I may propose, is to check whether the cached client uses same credential as incoming one. If yes, then use the cached client, if no then refresh the client.
This assumes that we can get credentials from client. if no, then this approach is not going to work.

tiwarishub: getting credentials from client is not there but i thought of storing PAT as static member in factory and for every getClient comparing the PAT with incoming PAT but this way i am introducing a dependency for example if in future if there is other param as well on which client is getting dependent then that param also we need to take care.

For 1st point, generally we call getClient function to get the client before using it, and it will take care giving proper client , if someone has used resetClient before getclient by mistake. But if some other class will also caching the client after getClient , then if by mistake they call resetClient and used the cached client then it will be an issue. But in that case that will be wrong approach to cache already cache client (which is static as well )

bishal-pdMSFT: > i thought of storing PAT as static member in factory and for every getClient comparing the PAT with incoming PAT but this way i am introducing a dependency for example if in future if there is other param come on which client is getting dependent then that param also we need to take care.

Agree

bishal-pdMSFT: I have one more suggestion. My primary concern is with exposing explicit function to reset client which seems like an unnecessary extra functionality to consumers. 
What consumer wants is ability to get a client which is refreshed (i.e. not the cached one). If so, then the facory's `getClient()` function can have an additional argument `refresh` which will mean that client wants a uncached client,

bishal-pdMSFT: Also, just looking at `getClient` again,  if the client is sending a `githubPatToken` then it implicitly means that it wants a uncached client. And we should honour that. So, IMO a solution would be make `githubPatToken` as optional argument and use cached client if PAT is not specified and refresh the client if it is specified.
Thoughts?

tiwarishub: Agree bishal, i will make a change to make service credentials and githubPAT as optional arguments ( because which credentials to take depends upon serviceframework option) and if anyone pass both then only we will create a new client and for rest cases we will return cached one or if cache is not there then will throw an exception

src/configure/clients/TemplateServiceClientFactory.ts
```diff
@@ -1,28 +1,43 @@
...
 
 export class TemplateServiceClientFactory {
 
-    public static async getClient(credentials?: ServiceClientCredentials, githubPatToken?: string): Promise<ITemplateServiceClient> {
+    public static async getClient(): Promise<ITemplateServiceClient> {
         if (!!this.client) {
             return this.client;
         }
+
+        if (!this.githubPatToken || !this.credentials) {
```
tiwarishub, line 15: Even though githubPat is not used for VSSF service framework, i have kept this githubPAT check condition because template service client will be only used  for github repos and there we will have a valid PAT and for azure-repo we use local templates

src/configure/clients/TemplateServiceClientFactory.ts
```diff
@@ -1,28 +1,43 @@
...
+            this.client = new TemplateServiceClient(serviceDefinition.serviceUrl, new TokenCredentials(this.githubPatToken, "token"), {
                 "User-Agent": UserAgent,
                 "Content-Type": "application/json; charset=utf-8"
             });
         }
+
         return this.client;
     }
 
+    public static setClientCredentials(credentials: ServiceClientCredentials, githubPatToken: string): void {
```
bishal-pdMSFT, line 34: rename to `initialize`. This way it is easier for someone to guess that they first need to call this function before getting client.

src/configure/configure.ts
```diff
@@ -203,6 +205,8 @@ class Orchestrator {
 
             const repoAnalysisResult = await this.getRepositoryAnalysis();
 
+            TemplateServiceClientFactory.setClientCredentials(this.inputs.azureSession.credentials, this.inputs.githubPATToken);
+            ProvisioningServiceClientFactory.setClientCredentials(this.inputs.azureSession.credentials, this.inputs.githubPATToken);
```
bishal-pdMSFT, line 209: move to a separate function named `initializeClients`. This makes it clear to understand that we are initizlizing client at this point in code flow.

bishal-pdMSFT (approved):

src/configure/clients/TemplateServiceClientFactory.ts
tiwarishub: done

src/configure/configure.ts
tiwarishub: done

PR #205, opened 2021-02-05
More Tolerant Conditions for Visibility Rules

mindlessroman:
Visibility rules for the the task.json (ie [when building a custom AzDO task](https://docs.microsoft.com/en-us/azure/devops/extend/develop/add-build-task?view=azure-devops#taskjson)) currently only allow for alphanumeric characters and spaces. 

This is very limiting if you have a string on the other side of the conditional - `someVariable = my_value` or `someVariableName != some-guid-goes-here` . Proposing to include hyphens and underscores to be more inclusive and therefore more flexible for options. The left side of the operation can stay alphanumeric-and-space, but the contents on the right side should be more flexible than it currently is.

See this issue in azure-pipelines-task-lib: https://github.com/microsoft/azure-pipelines-task-lib/issues/658 Someone there requested `#` to be a viable character but may be something to consider later?

### DIFF
--- a/src/configure/templateInputHelper/utilities/VisibilityHelper.ts
+++ b/src/configure/templateInputHelper/utilities/VisibilityHelper.ts
@@ -60,7 +60,7 @@ export class VisibilityHelper {
     }
 
     private static getPredicateRule(visibleRule: string): IPredicate {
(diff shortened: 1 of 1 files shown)


mindlessroman:
And it seems I had a bit of a brainfart - you all just maintain the code for the extension. I hope still that this can spark a discussion, maybe who to talk to.

bishal-pdMSFT:
@mindlessroman this repo is a vs code extension and not ADO tasks repo. I would suggest you to use Developer community forum to raise this issue or may be even try the tasks repo.

PR #206, opened 2021-02-25
Updated python for scm build at run time

tiwarishub:


src/configure/templates/githubWorkflowTemplates/pythonLinuxFunctionApp.yml
```diff
@@ -34,6 +34,12 @@ jobs:
         source worker_venv/bin/activate
         pip install -r {{{repositoryAnalysisApplicationSettings.settings.pythonRequirementsFilePath}}}
 
+    - name: 'Update app service setting'
+      uses: azure/appservice-settings@v1
+      with:
+        app-name: {{{ targetResource.resource.name }}}
+        app-settings-json: '[{ "name": "SCM_DO_BUILD_DURING_DEPLOYMENT", "value": "true" }]'
```
bishal-pdMSFT, line 41: @tiwarishub I notice few things which are different from what Jason had suggested in his [doc](https://microsoft-my.sharepoint.com/:w:/p/jafreebe/Ebv_PjO1P71DpvebaAVziYcBbeMvMiEOcgwk_zZXPPGZBA?e=EpmWQD&CID=36B1937C-10B2-44D2-80F3-FC48380DFA76&wdLOR=c503EB82D-5880-4D1C-A9C8-2F3A68AC5587). This doc refers [these ](https://gist.github.com/JasonFreeberg/c8911383f322797a5328cdb9913f0bc3) gists. This doc was mentioned in his email.
1. SCM_DO_BUILD_DURING_DEPLOYMENT is set in the deploy ( `functions-action`) step itself rather than as a separate step
2. It uses two jobs and pass the artifact from build phase by execluding virtual environment directory. This ensures that the package that deploy action uses does not have virtual env and that is expected as that would be created by Oryx build in server side.

bishal-pdMSFT (approved):

src/configure/templates/githubWorkflowTemplates/pythonLinuxFunctionApp.yml
tiwarishub: done

PR #212, opened 2021-03-31
Node function fix

tiwarishub:
Adding a check for template type as well to fix the issues like Node functions which as per of earlier condition going to provisioning service. but it should go local provisioning as the functions for the node language is not supported in the template service

Another fix includes is ``npm install`` fails  with the below error
```
Error installing vscode.d.ts: Error: Request returned status code: 404
Details: <!DOCTYPE html>
<html lang="en">
<head>
<meta charset="utf-8">
<title>Error</title>
</head>
<body>
<pre>Cannot GET /api/releases/stable</pre>
</body>
</html>
```
The quick fix for this is updating vscode module to its latest version. but also found that vscode module is [deprecated](https://github.com/Microsoft/vscode-extension-vscode#%EF%B8%8F-deprecated-use-typesvscode-and-vscode-test-instead-%EF%B8%8F) So later on, if want, we can move to ``@types/vscode`` and`` vscode-test``

### DIFF
--- a/src/configure/configure.ts
+++ b/src/configure/configure.ts
@@ -117,8 +117,7 @@ class Orchestrator {
 
     private doesLanguageAndTargetSupportRemoteProvisioning(): boolean {
         return extensionVariables.remoteConfigurerEnabled === true && this.inputs.sourceRepository.repositoryProvider === RepositoryProvider.Github &&
-            (this.inputs.targetResource.resource.type === TargetResourceType.AKS || (this.inputs.pipelineConfiguration.template.language === "Node" && this.inputs.targetResource.resource.type === TargetResourceType.WebApp)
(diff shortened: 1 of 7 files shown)


src/configure/configure.ts
```diff
@@ -117,7 +117,7 @@ class Orchestrator {
 
     private doesLanguageAndTargetSupportRemoteProvisioning(): boolean {
         return extensionVariables.remoteConfigurerEnabled === true && this.inputs.sourceRepository.repositoryProvider === RepositoryProvider.Github &&
-            (this.inputs.targetResource.resource.type === TargetResourceType.AKS || this.inputs.targetResource.resource.type === TargetResourceType.WebApp) && !!this.inputs.sourceRepository.remoteUrl;
+            (this.inputs.targetResource.resource.type === TargetResourceType.AKS || this.inputs.targetResource.resource.type === TargetResourceType.WebApp) && !!this.inputs.sourceRepository.remoteUrl && this.inputs.pipelineConfiguration.template.templateType === TemplateType.REMOTE;
```
kanika1894, line 120: As disussed, put appropriate comments.

kanika1894 (approved):

PR #240, opened 2022-03-11
Add deprecation announcement

bishal-pdMSFT:
Adding deprecation in README as we plan to deprecate the extension.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,9 +5,9 @@
 [Get it on the Visual Studio Code Marketplace!](https://marketplace.visualstudio.com/items?itemName=ms-vscode-deploy-azure.azure-deploy)
 
 
-## 📢 Attention - Deprecation notice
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -4,6 +4,14 @@
 
 [Get it on the Visual Studio Code Marketplace!](https://marketplace.visualstudio.com/items?itemName=ms-vscode-deploy-azure.azure-deploy)
 
+
+## 📢 Attention - Deprecation notice
```
ashwinsangem, line 8: image
Is there a way to highlight it in red to catch the eye? The content looks good to me.

ashwinsangem (approved):
LGTM

README.md
bishal-pdMSFT: Could not find a quick way to add color. I have increased the font, changed to CAPS and added one more emoji ⛔

